Issue #3, opened 2022-03-15
Design topic collections


jiazhang0:
This issue collects the topics about the detailed sub-system/module designs for enclave-cc. The high level arch design will add a section to refer to these topics. Please feel free to contribute your topic.

- [x] new component: agent enclave (formerly stub enclave) 
- [x] shim-rune changes @haosanzi 
- [x] rune and PAL API v4 changes @YangLiang3 
- [x] Local Attestation protocol for FUSE key transmission @bigdata-memory 
- [ ] FUSE encryption system

### References

- Development plan: https://github.com/confidential-containers/enclave-cc/issues/2
- High level design doc: https://github.com/confidential-containers/enclave-cc/pull/1

jiazhang0:
Add a new item "Local Attestation protocol for FUSE key transmission" assigned to @bigdata-memory

mythi:
> Add a new item "Local Attestation protocol for FUSE key transmission" assigned to @bigdata-memory

Do you think a separate key is needed? I was wondering would a seal key with the mrsigner policy be sufficient.

jiazhang0:
> > Add a new item "Local Attestation protocol for FUSE key transmission" assigned to @bigdata-memory
> 
> Do you think a separate key is needed? I was wondering would a seal key with the mrsigner policy be sufficient.

The so-called Local Attestation protocol include several request/response interactions to exchange parameters bound to local report for authentication against the possible MITM attack. For example, inspiring ECDH used by sgx sdk LA sample would be a choice. Of course, sealing key is another option. However, the implementation of negotiation process for sealing key is complex than ECDH and it is not hardware-agnostic. This is the details of EGETKEY(key_id="SEAL_KEY") according to SDM vol 3:

<img width="1436" alt="屏幕快照 2022-06-17 上午11 02 19" src="https://user-images.githubusercontent.com/10241344/174216546-a0129c7e-941b-4d4e-9a46-6d412149f299.png">

To ensure the generated sealing key by agent enclave is generatible for peer app-enclaves which possibly host different LibOS, there are several constraints to be satisfied at least.
- unified ISVSVN and CONFIGSVN versioning scheme.
- same minimal-and-allowed-attributions

If EGETKEY() for FUSE key generation only occurs in agent-enclave side and then transmitted to app-enclave,  the benefits of EGETKEY() are degenerate. We would rather use ECDH directly for easy life.

Please correct me if I have something wrong. @bigdata-memory

[...]

PR #11, opened 2022-06-17
[Future Work] Adding LA design for Filesystem Decryption Key Provisioning

bigdata-memory:
This PR adds a local attestation protocol design for FUSE key provisioning.


docs/design.md
```diff
@@ -104,3 +104,14 @@ several ways according to policy:
 
 Then like the usual LibOS pattern it will call the container's entry point as though the container
 was a process unto itself (rather than where it now sits like a library in the LibOS PAL process).
+
+## Local Attestation (v2) for FUSE Key Provisioning
```
jiazhang0, line 108: I think v2 is not necessary here

jiazhang0 (changes requested):

docs/design.md
bigdata-memory: Fixed, thanks.

docs/design.md
```diff
@@ -104,3 +104,14 @@ several ways according to policy:
...
 Then like the usual LibOS pattern it will call the container's entry point as though the container
 was a process unto itself (rather than where it now sits like a library in the LibOS PAL process).
+
+## Local Attestation for FUSE Key Provisioning
+
+![EnclaveCC_LA.png](images/EnclaveCC_LA.png)
+
+In most scenarios, it is one-shot transactions used for single FUSE key provisioning. However, the LA protocol can easily be extended to provision multiple keys for more general purposes.
+To achieve this, it is necessary to add a key id to the second request message sent along G_a and app's report, with the ID being used to indicate which key is being requested.
+In addition to adding this ID, the generated nonce and derived AEK also need to be in the cache on both sides to keep the secure session alive which has been established in initial transaction.
```
mythi, line 114: What `AEK` stands for?  Can you clarify "secure session" in the context of `ttrpc`.

docs/design.md
```diff
@@ -104,3 +104,14 @@ several ways according to policy:
 
 Then like the usual LibOS pattern it will call the container's entry point as though the container
 was a process unto itself (rather than where it now sits like a library in the LibOS PAL process).
+
+## Local Attestation for FUSE Key Provisioning
```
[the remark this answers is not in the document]
jiazhang0: Agreed.

docs/design.md
```diff
@@ -104,3 +104,14 @@ several ways according to policy:
...
+
+![EnclaveCC_LA.png](images/EnclaveCC_LA.png)
+
+In most scenarios, it is one-shot transactions used for single FUSE key provisioning. However, the LA protocol can easily be extended to provision multiple keys for more general purposes.
```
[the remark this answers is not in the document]
jiazhang0: "FUSE Key" is ephemeral and generated by agent-enclave. It is only used one time for app container.

docs/design.md
bigdata-memory: The `AEK` stands for algorithm encryption key, it is a derived key used for AEAD.
Regarding the secure session, it is tied to `AEK` and has shorter or equal lifetime than `ttrpc` connection. Thanks.

docs/design.md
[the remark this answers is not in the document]
bigdata-memory: Fixed, thanks.

docs/design.md
jodh-intel: Added to the acronym list: https://github.com/confidential-containers/documentation/wiki/Acronyms#aek

docs/design.md
```diff
@@ -104,3 +104,14 @@ several ways according to policy:
 
 Then like the usual LibOS pattern it will call the container's entry point as though the container
 was a process unto itself (rather than where it now sits like a library in the LibOS PAL process).
+
+## Local Attestation for Filesystem Decryption Key Provisioning
+
+![EnclaveCC_LA.png](images/EnclaveCC_LA.png)
+
+In most scenarios, it is one-shot transactions used for single filesystem decryption key provisioning. However, the LA protocol can easily be extended to provision multiple keys for more general purposes.
+To achieve this, it is necessary to add a key id to the second request message sent along G_a and app's report, with the ID being used to indicate which key is being requested.
```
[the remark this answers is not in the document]
bigdata-memory: `G_a` is the DH pubkey generated inside app. enclave, thanks.

jiazhang0:
@bigdata-memory Could you merge the 3 commits into single one?

jiazhang0 (approved):

mythi:
> @mythi @jodh-intel Please ack this PR. Thanks.

LGTM but the diagram still talks about "FUSE".

mythi:
@bigdata-memory one more open question from my side: to my knowledge, `ttrpc` has no secure transport so is the comms between the two libOSes transparently secured?

bigdata-memory:
> > @mythi @jodh-intel Please ack this PR. Thanks.
> 
> LGTM but the diagram still talks about "FUSE".

Thanks for point it out, and will fix it up.

bigdata-memory:
> @bigdata-memory one more open question from my side: to my knowledge, `ttrpc` has no secure transport so is the comms between the two libOSes transparently secured?

Yes, there is no secure transport configured between ttrpc endpoints in our case, and the LibOSes need to validate and check integrity of what they have received. With regard to the secret in transit, it was designed to be wrapped by shared key.

mythi:
> > @bigdata-memory one more open question from my side: to my knowledge, `ttrpc` has no secure transport so is the comms between the two libOSes transparently secured?
> 
> Yes, there is no secure transport configured between ttrpc endpoints in our case, and the LibOSes need to validate and check integrity of what they have received. With regard to the secret in transit, it was designed to be wrapped by shared key.

I think we could add this sentence in the doc too.

docs/design.md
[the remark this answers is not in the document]
bigdata-memory: Yes, it doesn't. I remember a similar issue being raised and discussed at a past weekly enclave-cc meeting, however, we conclude that it won't be added at initial stage as agent enclave has prior knowledge of which key need to be provisioned to app enclave at that phase. @jiazhang0 please chime in if any, Thanks.

docs/design.md
```diff
@@ -104,3 +104,14 @@ several ways according to policy:
 
 Then like the usual LibOS pattern it will call the container's entry point as though the container
 was a process unto itself (rather than where it now sits like a library in the LibOS PAL process).
+
+## Local Attestation for Filesystem Decryption Key Provisioning
+
+![EnclaveCC_LA.png](images/EnclaveCC_LA.png)
```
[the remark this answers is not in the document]
bigdata-memory: We predetermined the algorithms for use with DH key exchange scheme for simplicity and performance at this stage. I think a TLS like full-fledged negotiation mechanism might be considered to adding if desired.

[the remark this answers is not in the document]
bigdata-memory: Yes, it could be derived in G6, however, we think it could be better to postpone this operation until it is needed with regard to security, performance and resource consumption, thanks.

bigdata-memory:
> > > @bigdata-memory one more open question from my side: to my knowledge, `ttrpc` has no secure transport so is the comms between the two libOSes transparently secured?
> > 
> > 
> > Yes, there is no secure transport configured between ttrpc endpoints in our case, and the LibOSes need to validate and check integrity of what they have received. With regard to the secret in transit, it was designed to be wrapped by shared key.
> 
> I think we could add this sentence in the doc too.

Done, thank you for this advice.

docs/design.md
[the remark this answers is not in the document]
kailun-qin: The predetermination is fine. But in this current design, it seems that there are already enough interactions to carry out an algo negotiation (i.e., the performance benefit is not that clear)?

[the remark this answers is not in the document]
kailun-qin: Thanks for the explanation!

docs/design.md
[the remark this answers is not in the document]
kailun-qin: Thanks! Yep, I understand that the agent enclave needs have prior knowledge - but is this relevant to the proposed protocol (i.e., it's about adding the key ID kind of design instead of the provision steps)? It's just a bit weird that we highlight this in wording but it's not even mentioned in the flow chart. It's also obscure from implementation perspective.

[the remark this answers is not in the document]
jiazhang0: > Thanks! Yep, I understand that the agent enclave needs have prior knowledge - but is this relevant to the proposed protocol (i.e., it's about adding the key ID kind of design instead of the provision steps)? It's just a bit weird that we highlight this in wording but it's not even mentioned in the flow chart. It's also obscure from implementation perspective.

Because whether the key id can be automatically enumerated or not is decided by the caller ttrpc service/client and their contexts. At this moment, the caller is not discussed yet. We hope to merge the protocol definition completed prior to the implementation. By then, we could make the final decision to add the key id into protocol or not. @bigdata-memory I suggest to add such a comment to explain why the key id is not mentioned in flow chart. Thanks!

docs/design.md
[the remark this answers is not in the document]
kailun-qin: Got it and good to know. Thanks!

docs/design.md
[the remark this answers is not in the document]
kailun-qin: I see, thanks for the context info. Make sense to me.

bigdata-memory:
> @bigdata-memory Could you merge the 3 commits into single one?

Done, thanks.

jiazhang0:
@mythi @kailun-qin @jodh-intel If you don't have further comments, I will merge this PR.

kailun-qin (commented):
LGTM, thanks!

mythi:
> @mythi @kailun-qin @jodh-intel If you don't have further comments, I will merge this PR.

I think this ask from you is still missing:
> @bigdata-memory I suggest to add such a comment to explain why the key id is not mentioned in flow chart.

dcmiddle:
> @mythi @kailun-qin @jodh-intel If you don't have further comments, I will merge this PR.

@jiazhang0  Sorry for late feedback.. I added my thoughts in #3 since that's where the sealing discussion had started.

bigdata-memory:
> > @mythi @kailun-qin @jodh-intel If you don't have further comments, I will merge this PR.
> 
> I think this ask from you is still missing:
> 
> > @bigdata-memory I suggest to add such a comment to explain why the key id is not mentioned in flow chart.

Done, Thank you for reminding me.

vijaydhanraj:
Thanks @bigdata-memory @jiazhang0 for the PR, few questions,
1. Based on the figure it looks like, the goal is to mutually attest the agent and app enclave. But without a trusted third party(TTP) or `MRenclave` how do you identify the app enclave? There is no unique id that is specific to the app enclave. Any other enclave running on the same host could pretend to be the app enclave.
2. If TTPs are being used, can you clarify which entity acts as the TTP?
3. Also, how do the agent and app enclaves connect to this TTP? Would both agent and app need to do a RA with the TTP?

YangLiang3:
> Thanks @bigdata-memory @jiazhang0 for the PR, few questions,
> 
> 1. Based on the figure it looks like, the goal is to mutually attest the agent and app enclave. But without a trusted third party(TTP) or `MRenclave` how do you identify the app enclave? There is no unique id that is specific to the app enclave. Any other enclave running on the same host could pretend to be the app enclave.
> 2. If TTPs are being used, can you clarify which entity acts as the TTP?
> 3. Also, how do the agent and app enclaves connect to this TTP? Would both agent and app need to do a RA with the TTP?

Hi Vijay,
To my understanding, the agent enclave has been attested by RA, so app enclave can trust the agent enclave.  And agent enclave also can verify the payload receiver and LA client together by ttrpc LA(Payload receiver and LA client will have a fixed mrenclave)

bigdata-memory:
> Thanks @bigdata-memory @jiazhang0 for the PR, few questions,
> 
> 1. Based on the figure it looks like, the goal is to mutually attest the agent and app enclave. But without a trusted third party(TTP) or `MRenclave` how do you identify the app enclave? There is no unique id that is specific to the app enclave. Any other enclave running on the same host could pretend to be the app enclave.
> 2. If TTPs are being used, can you clarify which entity acts as the TTP?
> 3. Also, how do the agent and app enclaves connect to this TTP? Would both agent and app need to do a RA with the TTP?

Thanks @YangLiang3, Let me add one more point for your reference. As using ECDHE algorithm, with the pub keys (G_a, G_b) the Agent can distinguish app enclave instance from others.

vijaydhanraj:
Thanks @YangLiang3. So how does the agent know the `MRenclave` of the app enclave? Who provides it this information? From the figure this is not clear to me. And also how does the app enclave know that the agent is remotely attested and is not a malicious enclave. 

From the report, we can check the `MRsigner` and can prove that the enclaves are indeed running on Intel SGX hardware but the identity verification is not clear to me without a TTP.

jiazhang0:
> Thanks @bigdata-memory @jiazhang0 for the PR, few questions,
> 
>     1. Based on the figure it looks like, the goal is to mutually attest the agent and app enclave. But without a trusted third party(TTP) or `MRenclave` how do you identify the app enclave? There is no unique id that is specific to the app enclave. Any other enclave running on the same host could pretend to be the app enclave.
> 
>     2. If TTPs are being used, can you clarify which entity acts as the TTP?
> 
>     3. Also, how do the agent and app enclaves connect to this TTP? Would both agent and app need to do a RA with the TTP?

The essential logic here is that the LibOS components are deployed already by CSP prior to the deployment of workload from application container image, so once RA is established triggered by the initial attestation in order to get the container image decryption key from remote relying party, the mrenclave and other associated evidence claims are sent to remote relying party to be authenticated. The remote relying party will check against mrenclave reflecting the integrity of deployed LibOS and associated configurations to see whether it is trusted. This model is already applied to VM-based (e.g, tdx and sev and so on) confidential containers. This approach will de-couple the deployment of LibOS and app, and mrenclave will not reflect the integrity of app. The app itself is protected by container ecosystem security tools like container image signing and encryption scheme. In summary, the agent-enclave is unverified until the initial remote attestation occurs and then the relying party successfully authenticates it. After that agnet-enclave will decrypt and unpack the encrypted container image and then transfer the filesystem encryption key over LA channel. During the handshake of establishment the LA channel, app-enclave will additionally verify the identity of agent-enclave. I think you are focusing on the last piece of the whole story right? @vijaydhanraj

vijaydhanraj:
>During the handshake of establishment the LA channel, app-enclave will additionally verify the identity of agent-enclave

Yes, how does app-enclave verify the agent-enclave? Will a trusted third party give the agent's `MRenclave` to the app-enclave so that it can compare against the report?

Also, another thing: From the proposed design there is nothing preventing a malicious app-enclave running on the same hardware to start the attestation/handshake mechanism to get the fuse key. Is this possible?
Is there a trusted third party that provides the `MRenclave` of the app-enclave to the agent-enclave?

bigdata-memory:
@vijaydhanraj I have updated the document to address your concern you raised in meeting. thanks.

vijaydhanraj:
Thanks @bigdata-memory. Could you also please update the figure so that it becomes more clear?

bigdata-memory:
> Thanks @bigdata-memory. Could you also please update the figure so that it becomes more clear?

This has been covered as the steps of `check identity against policy` on both sides in the figure, Thanks.

vijaydhanraj:
This is very ambiguous and while reviewing I had many questions. If this is not explicitly captured it will lead to confusion in the future. So as discussed can you please add some text and update the figure saying there will be a trusted third party that validates the `MRenclave` of the other party? or whatever is the policy that we plan to incorporate to attest the enclaves between each other.

docs/design.md
```diff
@@ -104,3 +104,17 @@ several ways according to policy:
 
 Then like the usual LibOS pattern it will call the container's entry point as though the container
 was a process unto itself (rather than where it now sits like a library in the LibOS PAL process).
+
+## Local Attestation for Filesystem Decryption Key Provisioning
```
dcmiddle, line 108: I think the best path for enclave-cc right now is to use sealing. There is some discussion in #3 subsequent to the creation of this PR to support the sealing direction.

Please update the heading to reflect that Local Attestation is exploratory or future work.

You will also want to update the design to discuss the policy. The text in the diagram of "check [agent|app] identity against policy" isn't sufficient to evaluate the design here. 

It is also important to define the APIs. 

I would also appreciate some discussion of the use of network interfaces, i.e. are these remotely accessible? How does the app-enclave address the agent-enclave?

dcmiddle (changes requested):

jiazhang0:
> This is very ambiguous and while reviewing I had many questions. If this is not explicitly captured it will lead to confusion in the future. So as discussed can you please add some text and update the figure saying there will be a trusted third party that validates the `MRenclave` of the other party? or whatever is the policy that we plan to incorporate to attest the enclaves between each other.

The detailed method to archive identity verification is an implementation-specific so it is not documented in this PR. I think it is reasonable because this design focuses on protocol itself not covering the engineering details and how it will be implemented.

For me, a possible way to verify the peer identity is to use the mrenclave, assuming app-enclave and agent-enclave are hosted by the same version of libos and thus the mrenclave value are same between peers. Another possible way may be TTP as you mentioned. I think this info can be added when the implementation of this protocol is submitted by then. Do it make sense to you?

docs/design.md
jiazhang0: I think it would be fine to mark it as exploratory or future work. This doc just gives a design using TEE-agnostic approach to implement an attest channel to transfer a secret. Your questions are reasonable and could be addressed by the real implementation of this LA protocol. The discussion context is here: https://github.com/confidential-containers/enclave-cc/issues/3#issuecomment-1214284752

dcmiddle: In general, the additional detail I'm requesting is to distinguish the enclave-cc design from local attestation generally.

dcmiddle:
Please word wrap at 100 characters.

vijaydhanraj:
>detailed method to archive identity verification is an implementation-specific so it is not documented in this PR.

No, not a detailed method but at least a high-level policy for identification of the other enclave should be incorporated into the design. In attestation evaluating the identity of the other enclave is key and just saying `check agent/app against policy` feels a little insufficient. This is what I wanted to highlight.

>For me, a possible way to verify the peer identity is to use the mrenclave, assuming app-enclave and agent-enclave are hosted by the same version of libos and thus the mrenclave value are same between peers

Don't get this fully as even with same version of libos `agent-encalve` and `app-enclave` will have different code flows resulting in different `mrenclave` and not sure how `mrenclave` can be used. Or did you mean `mrsigner`?

bigdata-memory:
> 

Done, thanks.

jiazhang0:
> > detailed method to archive identity verification is an implementation-specific so it is not documented in this PR.
> 
> No, not a detailed method but at least a high-level policy for identification of the other enclave should be incorporated into the design. In attestation evaluating the identity of the other enclave is key and just saying `check agent/app against policy` feels a little insufficient. This is what I wanted to highlight.
> 
> > For me, a possible way to verify the peer identity is to use the mrenclave, assuming app-enclave and agent-enclave are hosted by the same version of libos and thus the mrenclave value are same between peers
> 
> Don't get this fully as even with same version of libos `agent-encalve` and `app-enclave` will have different code flows resulting in different `mrenclave` and not sure how `mrenclave` can be used. Or did you mean `mrsigner`?

I meant mrenclave exactly. This discussion happens in https://github.com/confidential-containers/enclave-cc/issues/3#issuecomment-1214284752

vijaydhanraj:
Thanks for the pointer. I have responded in that thread.

mythi:
> > detailed method to archive identity verification is an implementation-specific so it is not documented in this PR.
> 
> No, not a detailed method but at least a high-level policy for identification of the other enclave should be incorporated into the design. In attestation evaluating the identity of the other enclave is key and just saying `check agent/app against policy` feels a little insufficient. This is what I wanted to highlight.

+1, I don't think we are in a rush to get this merged until these design opens are clarified. It sounds the app enclave would have to provide additional info in the report identifying the container rootfs it tries to access and the agent enclave instance would only accept it if the container rootfs matches to what it has processed.

docs/design.md
```diff
@@ -104,3 +104,42 @@ several ways according to policy:
...
+Furthermore, agent or app. should set the expiration time of its own security context properly to
+minimize risk exposure. Upon closing of underlying ttrpc connection, the security context will be
+cleared accordingly.
+
+Please note that there is no secure transport configured between ttrpc endpoints in our case,
+therefore the LibOSes need to validate and check integrity of whatever they received. The ttrpc
+service needs to be strictly configured to listen on loopback interface only for security. With
+regard to the secret in transit, it was designed to be wrapped by a shared key. In addition, the
+identity of agent or app. could also be verified to further ensure the provenance of the app.
+enclave container.
```
dcmiddle, line 145: In the diagram please consistently name the FUSE key. Sometimes it is just mentioned as `key` which can be ambiguous in a key exchange protocol.

can you improve the wording in this step...
image

Something like `Encrypt` the fuse key using the AEK rather than `wrap`. 
Add Nonce 1 in `Associated Data`

docs/design.md
```diff
@@ -104,3 +104,42 @@ several ways according to policy:
...
+subsequent key provisioning requests.
+
+Furthermore, agent or app. should set the expiration time of its own security context properly to
+minimize risk exposure. Upon closing of underlying ttrpc connection, the security context will be
+cleared accordingly.
+
+Please note that there is no secure transport configured between ttrpc endpoints in our case,
+therefore the LibOSes need to validate and check integrity of whatever they received. The ttrpc
+service needs to be strictly configured to listen on loopback interface only for security. With
+regard to the secret in transit, it was designed to be wrapped by a shared key. In addition, the
```
dcmiddle, line 143: This sentence is unclear. "... it was designed to be wrapped by a shared key." 
Isn't that the whole point of this design? I'm not clear if something additional is intended by this sentence.

docs/design.md
```diff
@@ -104,3 +104,42 @@ several ways according to policy:
...
+identity of agent or app. could also be verified to further ensure the provenance of the app.
```
dcmiddle, line 144: You are already specifying that the app-enclave identity is explicitly checked by the agent-enclave.
In this sentence do you want to say that the compliment is also possible (i.e. app checks agent)?
If you want to say that please comment to what benefit that provides or risk that would address. It becomes very difficult to have two enclaves check separate MRENCLAVE values.

dcmiddle (changes requested):

docs/design.md
```diff
@@ -104,3 +104,42 @@ several ways according to policy:
...
+both sides to keep the secure session alive which has been established in initial transaction. The
+session's security context will hold those cached values to ensure AIC triad and freshness for
+subsequent key provisioning requests.
+
+Furthermore, agent or app. should set the expiration time of its own security context properly to
+minimize risk exposure. Upon closing of underlying ttrpc connection, the security context will be
+cleared accordingly.
+
+Please note that there is no secure transport configured between ttrpc endpoints in our case,
+therefore the LibOSes need to validate and check integrity of whatever they received. The ttrpc
```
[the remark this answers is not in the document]
bigdata-memory: added a link for reference, thanks.

docs/design.md
bigdata-memory: rephrased it and other associated parts, thanks.

docs/design.md
```diff
@@ -104,3 +104,42 @@ several ways according to policy:
...
+general purposes.
+
+To achieve this, it is necessary to add a key id to the second request message sent along G_a and
+app's report, with the ID being used to indicate which key is being requested. But whether the key
+id can be automatically enumerated or not is decided by the caller ttrpc service/client and their
+contexts. At this moment, the caller is not discussed yet. We hope to merge the protocol definition
+completed prior to the implementation. By then, we could make the final decision to add the key id
+into protocol or not.
+
+In addition to adding this ID, the generated nonce and derived AEK also expect to be in the cache on
```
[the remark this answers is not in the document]
bigdata-memory: Made it more specific, thanks.

docs/design.md
bigdata-memory: removed ambiguous part, thanks.

docs/design.md
bigdata-memory: removed the ambiguous part, thanks.

docs/design.md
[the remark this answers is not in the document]
dcmiddle: It is not clear if you are saying that the ttrpc limitation means that there are additional integrity precautions for the enclaves to satisfy (e.g. for the messages in this protocol), or if you are merely saying that we should use AEAD as specified to protect the FDK because AEAD provides integrity.

docs/design.md
```diff
@@ -104,3 +104,41 @@ several ways according to policy:
...
+The session's security context will hold those cached values to ensure AIC triad and freshness for
+subsequent key provisioning requests.
+
+Furthermore, agent or app. should set the expiration time of its own security context properly to
+minimize risk exposure. Upon closing of underlying `ttrpc` connection, the security context will be
+cleared accordingly.
+
+Please note that there is no secure transport configured between `ttrpc` endpoints in our case,
+therefore it is necessary to check the integrity and ensure the confidentiality if any of whatever
+they received (e.g., `AEAD`, https://datatracker.ietf.org/doc/html/rfc5116). The `ttrpc` service
```
dcmiddle, line 142: It is not clear if you are saying that the ttrpc limitation means that there are additional integrity precautions for the enclaves to satisfy (e.g. for the messages in this protocol), or if you are merely saying that we should use AEAD as specified to protect the FDK because AEAD provides integrity.

docs/design.md
```diff
@@ -104,3 +104,41 @@ several ways according to policy:
...
+needs to be strictly configured to listen on loopback interface only for security. In addition, the
+identity of app. could also be verified to ensure the provenance of the app. enclave.
```
dcmiddle, line 144: In this last sentence, I think 'app' and 'agent' got swapped on your latest revision.

But probably this all should be introduced early on and explicitly...
In your initial paragraphs you probably want to say that the agent holds a secret (FDK) and the app wants to authenticate to the agent in order for the agent to release the FDK. The app will prove its identity with a report and the agent will check that against a policy which requires a matching MRSIGNER or MRENCLAVE, possibly including an additional value from KSS. This policy is also really the heart of what this design should explain for enclave-cc.
(i.e. Local attestation as a generic protocol is shown in the linux-sgx examples and the SGX documentation, so discussing things specific to enclave-cc is most helpful for enclave-cc documentation and saves you trouble of rewriting things that may be documented elsewhere. 
Similarly the AEAD algorithm selection should be specified, but if the thinking is just to build on the linux-sgx example it might be better to just reference that.).


In this last sentence I think you meant to offer that the App could also authenticate the Agent. If you want to suggest that, please offer some discussion why this would be beneficial. The only things that come to mind for me are availability attacks which are outside the CC model.
Please also update the diagram to mark the app checking the agent against policy as optional or consider dropping that part.

docs/design.md
bigdata-memory: It is a design choice not the limitation of ttrpc, thanks.

docs/design.md
bigdata-memory: Thank you for the comments, I have updated the doc and diagram accordingly.

docs/design.md
```diff
@@ -104,3 +104,47 @@ several ways according to policy:
...
+The session's security context will hold those cached values to ensure AIC triad and freshness for
+subsequent key provisioning requests.
+
+Furthermore, agent or app. should set the expiration time of its own security context properly to
+minimize risk exposure. Upon closing of underlying `ttrpc` connection, the security context will be
+cleared accordingly.
+
+Please note that there is no secure transport configured between `ttrpc` endpoints in our case,
+therefore it is necessary to check the integrity and ensure the confidentiality if any of whatever
+they received using `AEAD`. Furthermore, the `ttrpc` service needs to be strictly configured to
```
dcmiddle, line 149: I think you just want to call out protecting the FDK specifically. If there are other messages you are trying to cover please add those or describe further what you mean. Thanks for all the updates.
```suggestion
therefore we rely on the confidentiality and integrity measures of `AEAD` to protect the FDK 
over this transport. Furthermore, the `ttrpc` service needs to be strictly configured to
```

bigdata-memory: Thank you very much for the commit suggestion.
I think there is no other coverage,
so there is no further description for now, thanks.

dcmiddle (approved):

docs/design.md
```diff
@@ -104,3 +104,47 @@ several ways according to policy:
...
+In this design, we leverages the `AEAD` algorithm for FDK protection over `ttrpc` channel. It is a
+recommended cryptographic algorithm in IETF, see [rfc5116](https://datatracker.ietf.org/doc/html/rfc5116).
+Regarding the ephemeral Diffie-Hellman key exchange, we chooses the Curve25519, which is recommended
+by IETF [rfc8031](https://www.rfc-editor.org/rfc/rfc8031). The design doesn't include the agent
+authentication in enclave-CC security context, however, the app can easily add app side user defined
+policy to enhance the security configuration.
+
+In most scenarios, it is one-shot transactions used for single filesystem decryption key (FDK)
+provisioning. However, this LA protocol can easily be extended to provision multiple keys for more
+general purposes.
```
mythi, line 129:
```suggestion
general purposes or, for example, multiple application containers in the same pod.
```

docs/design.md
```diff
@@ -104,3 +104,47 @@ several ways according to policy:
...
+developers to implement this feature.
+
+In this design, we leverages the `AEAD` algorithm for FDK protection over `ttrpc` channel. It is a
+recommended cryptographic algorithm in IETF, see [rfc5116](https://datatracker.ietf.org/doc/html/rfc5116).
+Regarding the ephemeral Diffie-Hellman key exchange, we chooses the Curve25519, which is recommended
+by IETF [rfc8031](https://www.rfc-editor.org/rfc/rfc8031). The design doesn't include the agent
+authentication in enclave-CC security context, however, the app can easily add app side user defined
+policy to enhance the security configuration.
+
+In most scenarios, it is one-shot transactions used for single filesystem decryption key (FDK)
```
mythi, line 127:
```suggestion
In most scenarios, it is one-shot transactions used for single FDK

File: docs/design.md [hunk 18]
```diff
@@ -104,3 +104,47 @@ several ways according to policy:
...
+provisioning. However, this LA protocol can easily be extended to provision multiple keys for more
+general purposes.
+
+To achieve this, it is necessary to add a key id to the second request message sent along G_a and
+app's report, with the ID being used to indicate which key is being requested. But whether the key
```
mythi, line 132:
```suggestion
app's report, with the key id being used to indicate which key is being requested. But whether the key
```

mythi: it looks `key id` is commonly used in this paragraph

docs/design.md
```diff
@@ -104,3 +104,47 @@ several ways according to policy:
...
 was a process unto itself (rather than where it now sits like a library in the LibOS PAL process).
+
+## [Future Work] Local Attestation for Filesystem Decryption Key Provisioning
+
+![EnclaveCC_LA.png](images/EnclaveCC_LA.png)
+
+A local attestation (LA) feature is needed for authentication/authorization and secret provisioning
+between an Agent-enclave instance and a set of App-enclave instances resided in a same Pod. The
+agent holds a secret for filesystem decryption and the app wants to authenticate to the agent in
+order for the agent to release the filesystem decryption key (FDK). The app will prove its identity
```
mythi, line 115:
```suggestion
order for the agent to release the filesystem decryption key (FDK). The app-enclave will prove its identity
```

mythi (commented):
mostly minor editorial comments with the goal of making some sentences easier to read

LGTM

docs/design.md
bigdata-memory: Agree, thanks.

docs/design.md
bigdata-memory: Thank you so much for those commit suggestions.
It is very helpful to improve the readability. Thanks.

bigdata-memory:
Thanks @hdxia for comments on this diagram.

mythi (approved):

mythi:
> Thanks @hdxia for comments on this diagram.

Are they available somewhere?

bigdata-memory:
> > Thanks @hdxia for comments on this diagram.
> 
> Are they available somewhere?

They are meeting comments so that's why no Github's comments here. thanks.

PR #14, opened 2022-06-30
enclave-agent: it is responsible for image management service.

YangLiang3:
1 As the ttrpc server, it receive requests sent by shim-rune.
2 Pull and unpack image.
3 Convert the image to the encrypted filesystem as the mount source of the app container.

Signed-off-by: Zhiwei Han <zhiwei1.han@intel.com>

### DIFF
--- a/src/enclave-agent/Cargo.toml
+++ b/src/enclave-agent/Cargo.toml
@@ -1,0 +1,32 @@
+[package]
+name = "enclave-agent"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+protobuf = { version = "=2.14.0" }
+bytes = "0.4.11"
+libc = "0.2.58"
+byteorder = "1.3.2"
+log = "0.4.11"
+simple-logging = "2.0.2"
+nix = "0.23.0"
+ttrpc = { git = "https://github.com/haosanzi/ttrpc-rust", features = ["async"] }
+ctrlc = { version = "3.0", features = ["termination"] }
+
+# Async runtime
+tokio = { version = "1.14.0", features = ["full"] }
+async-trait = "0.1.42"
+
+image-rs = { git = "https://github.com/zhiwei-intel-h/image-rs", branch = "dev-occlum-adaption", features = ["occlum_feature"], default-features = false }
+kata-sys-util = { git = "https://github.com/kata-containers/kata-containers" }
+
+anyhow = "1.0.32"
+clap = "2.33.3"
+
+protocols = { path = "../libs/protocols" }
+
+[build-dependencies]
+ttrpc-codegen = "0.2.0"
--- a/src/enclave-agent/Cargo.toml
+++ b/src/enclave-agent/Cargo.toml
@@ -20,8 +20,8 @@ ctrlc = { version = "3.0", features = ["termination"] }
 tokio = { version = "1.14.0", features = ["full"] }
 async-trait = "0.1.42"
 
-image-rs = { git = "https://github.com/zhiwei-intel-h/image-rs", branch = "dev-occlum-adaption", features = ["occlum_feature"], default-features = false }
-kata-sys-util = { git = "https://github.com/kata-containers/kata-containers" }
+image-rs = { git = "https://github.com/confidential-containers/image-rs", features = ["occlum_feature"], default-features = false }
+kata-sys-util = { git = "https://github.com/kata-containers/kata-containers", rev = "4b57c04c3379d6adc7f440d156f0e4c42ac157df" }
 
 anyhow = "1.0.32"
 clap = "2.33.3"
--- a/src/enclave-agent/src/main.rs
+++ b/src/enclave-agent/src/main.rs
@@ -1,0 +1,138 @@
+use std::env;
+use std::path::Path;
+use std::sync::Arc;
+
+use anyhow::{anyhow, Result};
+use async_trait::async_trait;
+use clap::{crate_authors, crate_version, App, Arg};
+use image_rs::config::ImageConfig;
+use image_rs::image::ImageClient;
+use image_rs::snapshots;
+use kata_sys_util::validate;
+use protocols::{image, image_ttrpc};
+use tokio::signal::unix::{signal, SignalKind};
+use tokio::sync::Mutex;
+use ttrpc::asynchronous::Server;
+use ttrpc::{self, error::get_rpc_status as ttrpc_error};
+
+const CONTAINER_BASE: &str = "/run/enclave-cc/containers";
+const CC_IMAGE_WORK_DIR: &str = "/run/image/";
+
+// TODO: will replace with unix socket
+const SOCK_ADDR: &str = "tcp://0.0.0.0:7788";
+
+struct ImageService {
+    image_client: Arc<Mutex<ImageClient>>,
+}
+
+impl ImageService {
+    fn new() -> Self {
+        env::set_var("CC_IMAGE_WORK_DIR", CC_IMAGE_WORK_DIR);
+        let new_config = ImageConfig {
+            default_snapshot: snapshots::SnapshotType::OcclumUnionfs,
+            ..Default::default()
+        };
+        Self {
+            image_client: Arc::new(Mutex::new(ImageClient {
+                config: new_config,
+                ..Default::default()
+            })),
+        }
+    }
+
+    async fn pull_image(&self, req: &image::PullImageRequest) -> Result<String> {
+        let image = req.get_image();
+        let mut cid = req.get_container_id().to_string();
+
+        if cid.is_empty() {
+            let v: Vec<&str> = image.rsplit('/').collect();
+            if !v[0].is_empty() {
+                // ':' have special meaning for umoci during upack
+                cid = v[0].replace(":", "_");
+            } else {
+                return Err(anyhow!("Invalid image name: {:?}", image));
+            }
+        } else {
+            validate::verify_id(&cid)?;
+        }
+
+        let source_creds = (!req.get_source_creds().is_empty()).then(|| req.get_source_creds());
+
+        let bundle_path = Path::new(CONTAINER_BASE).join(&cid);
+
+        self.image_client
+            .lock()
+            .await
+            .pull_image(image, &bundle_path, &source_creds, &None)
+            .await?;
+
+        Ok(image.to_owned())
+    }
+}
+
+#[async_trait]
+impl protocols::image_ttrpc::Image for ImageService {
+    async fn pull_image(
+        &self,
+        _ctx: &ttrpc::r#async::TtrpcContext,
+        req: image::PullImageRequest,
+    ) -> ttrpc::Result<image::PullImageResponse> {
+        match self.pull_image(&req).await {
+            Ok(r) => {
+                println!("Pull image {:?} successfully", r.clone());
+                let mut resp = image::PullImageResponse::new();
+                resp.image_ref = r;
+                return Ok(resp);
+            }
+            Err(e) => {
+                return Err(ttrpc_error(ttrpc::Code::INTERNAL, e.to_string()));
+            }
+        }
+    }
+}
+
+#[tokio::main(worker_threads = 1)]
+async fn main() {
+    let matches = App::new("Enclave agent")
+        .version(crate_version!())
+        .author(crate_authors!())
+        .arg(
+            Arg::with_name("listen")
+                .short("l")
+                .long("listen")
+                .value_name("sockaddr")
+                .help(
+                    ("Specify the socket listen addr. Default is ".to_string()
+                        + format!("{}", SOCK_ADDR).as_str())
+                    .as_str(),
+                )
+                .takes_value(true),
+        )
+        .get_matches();
+
+    let mut sockaddr = SOCK_ADDR;
+    if matches.is_present("listen") {
+        sockaddr = matches.value_of("listen").unwrap();
+    }
+
+    let image_service = Box::new(ImageService::new()) as Box<dyn image_ttrpc::Image + Send + Sync>;
+
+    let image_service = image_ttrpc::create_image(Arc::new(image_service));
+
+    let mut server = Server::new()
+        .bind(sockaddr)
+        .unwrap()
+        .register_service(image_service);
+
+    let mut interrupt = signal(SignalKind::interrupt()).unwrap();
+    server.start().await.unwrap();
+
+    println!("ttRPC server started: {:?}", sockaddr);
+
+    tokio::select! {
+        _ = interrupt.recv() => {
+            println!("shutdown the server");
+            server.shutdown().await.unwrap();
+        }
+    };
+}
--- a/src/enclave-agent/src/main.rs
+++ b/src/enclave-agent/src/main.rs
@@ -1,4 +1,7 @@
 use std::env;
+use std::fs;
+use std::fs::File;
+use std::io::prelude::*;
 use std::path::Path;
 use std::sync::Arc;
 
@@ -16,10 +19,9 @@ use ttrpc::asynchronous::Server;
 use ttrpc::{self, error::get_rpc_status as ttrpc_error};
 
 const CONTAINER_BASE: &str = "/run/enclave-cc/containers";
-const CC_IMAGE_WORK_DIR: &str = "/run/image/";
 
 // TODO: will replace with unix socket
-const SOCK_ADDR: &str = "tcp://0.0.0.0:7788";
+const TCP_SOCK_ADDR: &str = "tcp://0.0.0.0:7788";
 
 struct ImageService {
     image_client: Arc<Mutex<ImageClient>>,
@@ -27,7 +29,6 @@ struct ImageService {
 
 impl ImageService {
     fn new() -> Self {
-        env::set_var("CC_IMAGE_WORK_DIR", CC_IMAGE_WORK_DIR);
         let new_config = ImageConfig {
             default_snapshot: snapshots::SnapshotType::OcclumUnionfs,
             ..Default::default()
@@ -56,14 +57,48 @@ impl ImageService {
             validate::verify_id(&cid)?;
         }
 
+        let keyprovider_config = Path::new("/etc").join("ocicrypt_keyprovider_native.conf");
+        if !keyprovider_config.exists() {
+            let config = r#"
+            {
+                "key-providers": {
+                    "attestation-agent": {
+                        "native": "attestation-agent"
+                    }
+                }
+            }
+            "#;
+            File::create(&keyprovider_config)?.write_all(config.as_bytes())?;
+        }
+        std::env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", keyprovider_config);
+
+        let mut config: String = String::new();
+        let args: Vec<String> = env::args().collect();
+        // If config file specified in the args, read contents from config file
+        let config_position = args.iter().position(|a| a == "--decrypt-config" || a == "-c");
+        if let Some(config_position) = config_position {
+            if let Some(config_file) = args.get(config_position + 1) {
+                config = fs::read_to_string(config_file).expect("Config file not found");
+            } else {
+                panic!("The config argument wasn't formed properly: {:?}", args);
+            }
+        }
+
+        let decrypt_config = if !config.is_empty() {
+            Some(config.as_str())
+        } else {
+            None
+        };
+
         let source_creds = (!req.get_source_creds().is_empty()).then(|| req.get_source_creds());
 
         let bundle_path = Path::new(CONTAINER_BASE).join(&cid);
 
+        println!("Pulling {:?}", image);
         self.image_client
             .lock()
             .await
-            .pull_image(image, &bundle_path, &source_creds, &None)
+            .pull_image(image, &bundle_path, &source_creds, &decrypt_config)
             .await?;
 
         Ok(image.to_owned())
@@ -92,7 +127,7 @@ impl protocols::image_ttrpc::Image for ImageService {
 }
 
 #[tokio::main(worker_threads = 1)]
-async fn main() {
+async fn main() -> Result<()> {
     let matches = App::new("Enclave agent")
         .version(crate_version!())
         .author(crate_authors!())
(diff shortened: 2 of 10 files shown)


src/libs/protocols/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+default = []
+with-serde = [ "serde", "serde_json" ]
+
+[dependencies]
+#ttrpc = { version = "0.5.0", features = ["async"] }
```
jodh-intel, line 13: Commented out code.

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+// TODO: will replace with unix socket
+const SOCK_ADDR: &str = "tcp://0.0.0.0:7788";
+
+struct ImageService {
+    image_client: Arc<Mutex<ImageClient>>,
+}
+
+// A container ID must match this regex:
+//     ^[a-zA-Z0-9][a-zA-Z0-9_.-]+$
+fn verify_cid(id: &str) -> Result<()> {
```
jodh-intel, line 31: I seem to remember writing this code (and lots of unit tests) for the Kata Containers agent: see https://github.com/kata-containers/kata-containers/pull/1521 :smile:

It might be a good idea to mention this code was taken from the Kata Containers agent and include details of which commit the code was copied from in your commit.

src/enclave-agent/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+simple-logging = "2.0.2"
+nix = "0.23.0"
+ttrpc = { git = "https://github.com/haosanzi/ttrpc-rust", features = ["async"] }
+ctrlc = { version = "3.0", features = ["termination"] }
+
+# Async runtime
+tokio = { version = "1.14.0", features = ["full"] }
+async-trait = "0.1.42"
+
+image-rs = { git = "https://github.com/zhiwei-intel-h/image-rs", branch = "dev-occlum-adaption", features = ["unionfs_feature"] }
```
fidencio, line 23: Do you know when this will be merged into the `main` branch?
Not being on `main` is not a blocker, but I'd like to avoid depending on moving fork if possible.

src/libs/protocols/Cargo.toml
dcmiddle: related.. line below is pulling from a fork. Is there something required in there that isn't in upstream ttrpc?

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,152 @@
+use std::env;
+use std::path::Path;
+use std::sync::Arc;
+
+use clap::{crate_authors, crate_version, App, Arg};
+//use log::LevelFilter;
+use log::*;
```
dcmiddle, line 7: unused import warning

src/enclave-agent/src/main.rs
dcmiddle: Reflecting some verbal discussion with @jodh-intel ... For purposes of linking patches in future issues it might be best to put that reference in the code comments here (rather than the commit message). A longer term fix might be breaking this out as a separate module both projects can depend on.

YangLiang3: > I seem to remember writing this code (and lots of unit tests) for the Kata Containers agent: see [kata-containers/kata-containers#1521](https://github.com/kata-containers/kata-containers/pull/1521) 😄
> 
> It might be a good idea to mention this code was taken from the Kata Containers agent and include details of which commit the code was copied from in your commit.

Thanks James, we are modifying this issue.

src/libs/protocols/Cargo.toml
YangLiang3: > related.. line below is pulling from a fork. Is there something required in there that isn't in upstream ttrpc?

hi dan, current occlum can't support UDS. Offical ttrpc can't support tcp, so we add tcp support in ttrpc. Once the Occlum NGO can support UDS, we will use the upstream ttrpc

zhiwei-intel-h: > Commented out code.

Get it. Done

src/enclave-agent/src/main.rs
YangLiang3: > Reflecting some verbal discussion with @jodh-intel ... For purposes of linking patches in future issues it might be best to put that reference in the code comments here (rather than the commit message). A longer term fix might be breaking this out as a separate module both projects can depend on.

got it, thanks dan

zhiwei-intel-h:
> Please add unit tests to this code. https://github.com/confidential-containers/community/blob/main/PR-Review-Guide.md#testing [cc: @jodh-intel ]

Ok， I will.  Thanks

src/enclave-agent/src/main.rs
zhiwei-intel-h: Yes, fixed.

zhiwei-intel-h:
> Please add unit tests to this code. https://github.com/confidential-containers/community/blob/main/PR-Review-Guide.md#testing [cc: @jodh-intel ]

Currently, I can not add unit test for function pull_image on host envirenment, because pull_image depends on the specific Occlum envirenment.  Function pull_image will call image-rs crate which provide pull/unpack image and mount function, but the mount function have to depend on the implementation of Occlum.

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+use anyhow::{anyhow, Result};
+use image_rs::config::ImageConfig;
+use image_rs::image::ImageClient;
+use image_rs::snapshots;
+
+const CONTAINER_BASE: &str = "/run/enclave-cc-containers";
+const CC_IMAGE_WORK_DIR: &str = "/run/image/";
+
+// TODO: will replace with unix socket
+const SOCK_ADDR: &str = "tcp://0.0.0.0:7788";
```
dcmiddle, line 23: FYI, I created an issue for the TODO item. https://github.com/confidential-containers/enclave-cc/issues/18

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+use async_trait::async_trait;
+use tokio::signal::unix::{signal, SignalKind};
+use tokio::sync::Mutex;
+
+use anyhow::{anyhow, Result};
+use image_rs::config::ImageConfig;
+use image_rs::image::ImageClient;
+use image_rs::snapshots;
+
+const CONTAINER_BASE: &str = "/run/enclave-cc-containers";
```
jodh-intel, line 19: - This is quite a long path.
- If you split this into three components, we have more flexibility to  store additional state at some later stage. Something like `/run/enclave-cc/containers/` or maybe `/run/enclave-cc/ctrs/`.

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    match valid {
+        true => Ok(()),
+        false => Err(anyhow!("invalid container ID: {:?}", id)),
+    }
+}
+
+impl ImageService {
+    fn new() -> Self {
+        env::set_var("CC_IMAGE_WORK_DIR", CC_IMAGE_WORK_DIR);
```
jodh-intel, line 54: - What is this variable used by? Please document.
- Why is it set in `new()`? Why not just set it in a `setup()` function/method once?

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+        let image = req.get_image();
+        let mut cid = req.get_container_id().to_string();
+
+        if cid.is_empty() {
+            let v: Vec<&str> = image.rsplit('/').collect();
+            if !v[0].is_empty() {
+                // ':' have special meaning for umoci during upack
+                cid = v[0].replace(":", "_");
+            } else {
+                return Err(anyhow!("Invalid image name. {}", image));
```
jodh-intel, line 77: Nit: I'd quote the name for additional clarity.

```suggestion
                return Err(anyhow!("Invalid image name: {:?}", image));
```

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+async fn main() {
+    let matches = App::new("Enclave agent")
+        .version(crate_version!())
+        .author(crate_authors!())
+        .arg(
+            Arg::with_name("listen")
+                .short("l")
+                .long("listen")
+                .value_name("sockaddr")
+                .help("Specify the socket listen addr. Default is tcp://0.0.0.0:7788")
```
jodh-intel, line 129: Please don't hard-code values in strings - you could use `format!()` here to reference `SOCK_ADDR` to avoid doing that.

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+                return Ok(resp);
+            }
+            Err(e) => {
+                return Err(ttrpc_error(ttrpc::Code::INTERNAL, e.to_string()));
+            }
+        }
+    }
+}
+
+//#[tokio::main(flavor = "current_thread")]
```
jodh-intel, line 118: Commented out code.

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    let mut server = Server::new()
+        .bind(sockaddr)
+        .unwrap()
+        .register_service(image_service);
+
+    let mut interrupt = signal(SignalKind::interrupt()).unwrap();
+    server.start().await.unwrap();
+
+    println!("ttRPC server started");
```
jodh-intel, line 151: It might be helpful to also specify `SOCK_ADDR` in this log message.

src/libs/protocols/protos/image.proto
```diff
@@ -0,0 +1,25 @@
...
+package grpc;
+
+// Image defines the public APIs for managing images.
+service Image {
+    // PullImage pulls an image with authentication config.
+    rpc PullImage(PullImageRequest) returns (PullImageResponse) {}
+}
+
+message PullImageRequest {
+    // Image name (e.g. docker.io/library/busybox:latest).
```
jodh-intel, line 12: +1 for giving an example, but it would be useful to add a comment either explaining the full format of `image`, or providing a canonical URL for a doc that explains it. For example:

- Can I use an alternative to `docker.io`?
- is `:$tag` optional? If not specified, which version is pulled?
- `latest` is useful for users, but not helpful for problem determination (_"what was the value of 'latest' 2 weeks ago?"_) Hence, Is `latest` expanded to an actual version number and logged at pull time for auditing purposes?

jodh-intel (commented):
Thanks @YangLiang3 - a few comments...

src/enclave-agent/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+simple-logging = "2.0.2"
+nix = "0.23.0"
+ttrpc = { git = "https://github.com/haosanzi/ttrpc-rust", features = ["async"] }
+ctrlc = { version = "3.0", features = ["termination"] }
+
+# Async runtime
+tokio = { version = "1.14.0", features = ["full"] }
+async-trait = "0.1.42"
+
+image-rs = { git = "https://github.com/zhiwei-intel-h/image-rs", branch = "dev-occlum-adaption", features = ["occlum_feature"], default-features = false }
```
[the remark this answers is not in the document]
zhiwei-intel-h: > Same comment as above. We shouldn't be using personal clone of official projects for a number of reasons. It's also worth pointing out that these imports are potentially dangerous and/or fragile as they are not being pinned to a version number or a commit.

This code will be merged into  official project in the near future(https://github.com/confidential-containers/image-rs/pull/37). Once merged, the personal link will be replaced with official link.

src/enclave-agent/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[dependencies]
+protobuf = { version = "=2.14.0" }
+bytes = "0.4.11"
+libc = "0.2.58"
+byteorder = "1.3.2"
+log = "0.4.11"
+simple-logging = "2.0.2"
+nix = "0.23.0"
+ttrpc = { git = "https://github.com/haosanzi/ttrpc-rust", features = ["async"] }
```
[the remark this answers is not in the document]
zhiwei-intel-h: > Can we move to the official repo yet? If not, please can you raise a GitHub issue and add a link to it here explaining what is blocking the switch.

Yes, it is a known issue. https://github.com/confidential-containers/enclave-cc/issues/18

src/enclave-agent/src/main.rs
zhiwei-intel-h: Get it. Thanks.

src/enclave-agent/src/main.rs
zhiwei-intel-h: Get it. Thanks.

src/enclave-agent/src/main.rs
zhiwei-intel-h: Get it. Thanks

src/enclave-agent/src/main.rs
zhiwei-intel-h: Get it. Thanks.

src/enclave-agent/src/main.rs
zhiwei-intel-h: Get it. thanks

src/enclave-agent/src/main.rs
zhiwei-intel-h: > * What is this variable used by? Please document.
> * Why is it set in `new()`? Why not just set it in a `setup()` function/method once?

It refers to the implementation of kata https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/image_rpc.rs#L52

src/libs/protocols/protos/image.proto
zhiwei-intel-h: Ok, will add some link for reference to explaining the full format of docker image.

src/enclave-agent/Cargo.toml
zhiwei-intel-h: > Do you know when this will be merged into the `main` branch? Not being on `main` is not a blocker, but I'd like to avoid depending on moving fork if possible.

This personal project is temporary and will be merged into the official image-rs in the near future.

src/enclave-agent/src/main.rs
```diff
@@ -56,14 +59,47 @@ impl ImageService {
...
         }
 
+        let keyprovider_config = Path::new("/etc").join("ocicrypt_keyprovider_native.conf");
+        if !keyprovider_config.exists() {
+            let config = r#"
+            {
+                "key-providers": {
+                    "attestation-agent": {
+                        "native": "attestation-agent"
+                    }
```
xhuan28, line 69: Can the config be something like:
            {
                "key-providers": {
                    "attestation-agent": {
                        "native": "<kbc_name>:<kbs_uri>"
                    }
                }
            }
For example, in the eaa_kbc mode of attestation agent, it should be something like "eaa_kbc:127.0.0.1:1234"

zhiwei-intel-h: I think that KBC instance is not configed here while configed like this `provider:attestation-agent:eaa_kbc::127.0.0.1:1234` in decrypt_config.json

src/enclave-agent/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+ctrlc = { version = "3.0", features = ["termination"] }
+
+# Async runtime
+tokio = { version = "1.14.0", features = ["full"] }
+async-trait = "0.1.42"
+
+image-rs = { git = "https://github.com/zhiwei-intel-h/image-rs", branch = "dev-occlum-adaption", features = ["occlum_feature"], default-features = false }
```
haosanzi: Please use `https://github.com/confidential-containers/image-rs` since it has supported occlum unionfs

src/enclave-agent/src/main.rs
haosanzi, line 69: Why do we need to create keyprovider_config since kata-agent does not create this file?

src/enclave-agent/Cargo.toml
zhiwei-intel-h: > Please use `https://github.com/confidential-containers/image-rs` since it has supported occlum unionfs

I have replaced it with `image-rs = { git = "https://github.com/confidential-containers/image-rs", features = ["occlum_feature"], default-features = false }`.

dcmiddle (approved):

src/enclave-agent/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+nix = "0.23.0"
+ttrpc = { git = "https://github.com/haosanzi/ttrpc-rust", features = ["async"] }
+ctrlc = { version = "3.0", features = ["termination"] }
+
+# Async runtime
+tokio = { version = "1.14.0", features = ["full"] }
+async-trait = "0.1.42"
+
+image-rs = { git = "https://github.com/confidential-containers/image-rs", features = ["occlum_feature"], default-features = false }
+kata-sys-util = { git = "https://github.com/kata-containers/kata-containers" }
```
jodh-intel, line 24: Do you want to pull in `main:HEAD` or a versioned stable branch / tag?

src/enclave-agent/src/main.rs
jodh-intel: Since that's in a different project, I think it would be helpful to add that URL as a comment in the code.

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+        let args: Vec<String> = env::args().collect();
+        // If config file specified in the args, read contents from config file
+        let config_position = args.iter().position(|a| a == "--config" || a == "-c");
+        if let Some(config_position) = config_position {
+            if let Some(config_file) = args.get(config_position + 1) {
+                config = fs::read_to_string(config_file).expect("Config file not found");
+            } else {
+                panic!("The config argument wasn't formed properly: {:?}", args);
+            }
+        }
```
jodh-intel, line 87: Please can you use the `clap` routines (already added on this PR) for argument parsing.

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+        .version(crate_version!())
+        .author(crate_authors!())
+        .arg(
+            Arg::with_name("listen")
+                .short("l")
+                .long("listen")
+                .value_name("sockaddr")
+                .help(&format!(
+                    "{}{}",
+                    "Specify the socket listen addr. Default is ", SOCK_ADDR
```
jodh-intel, line 142: I know it's going to change, but I think it would still be worth specifying the valid types of socket that can be specified here.

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+                ))
+                .takes_value(true),
+        )
+        .arg(
+            Arg::with_name("decrypt config")
+                .short("c")
+                .long("config")
```
jodh-intel, line 149: Spaces in arg names seems potentially error-prone to me. Why not:

```suggestion
            Arg::with_name("decrypt-config")
                .short("c")
                .long("decrypt-config")
```

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+                .help(&format!("The decrypt config file path"))
+                .takes_value(true)
+                .required(false),
+        )
+        .get_matches();
+
+    let mut sockaddr = SOCK_ADDR;
+    if matches.is_present("listen") {
+        sockaddr = matches.value_of("listen").unwrap();
+    }
```
jodh-intel, line 159: You can avoid the `mut` and `unwrap()` using something like:

```suggestion
    let sockaddr = if let Some(addr) = matches.is_present("listen") {
        addr
    } else {
        SOCK_ADDR
    };
```

src/enclave-agent/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    let image_service = Box::new(ImageService::new()) as Box<dyn image_ttrpc::Image + Send + Sync>;
+
+    let image_service = image_ttrpc::create_image(Arc::new(image_service));
+
+    let mut server = Server::new()
+        .bind(sockaddr)
+        .unwrap()
```
jodh-intel, line 167: Rather than all the `unwrap()`'s, you could make `main()` return a `Result<()>`, or do the trick the Kata agent uses and have main be non-async, return a `Result` and `block_on()` an async `real_main() -> Result<()>`.

jodh-intel (commented):
Thanks @YangLiang3 - a few comments...

src/enclave-agent/src/main.rs
zhiwei-intel-h: > Spaces in arg names seems potentially error-prone to me. Why not:

Actually, we use `--config`  instead of `--decrypt config` to specify the config file. And `--decrypt-config` is also a good idea.

src/enclave-agent/src/main.rs
zhiwei-intel-h: Ok, get it.

src/enclave-agent/src/main.rs
zhiwei-intel-h: Ok, get it.

src/enclave-agent/src/main.rs
zhiwei-intel-h: > I know it's going to change, but I think it would still be worth specifying the valid types of socket that can be specified here.

Ok, I will change the SOCK_ADDR to TCP_SOCK_ADDR.

zhiwei-intel-h: > I know it's going to change, but I think it would still be worth specifying the valid types of socket that can be specified here.

Ok, I will change `SOCK_ADDR` to `TCP_SOCK_ADDR`.

src/enclave-agent/Cargo.toml
zhiwei-intel-h: I will specify a commit id to pull the crate `kata-sys-util`.

src/enclave-agent/src/main.rs
zhiwei-intel-h: > Please can you use the `clap` routines (already added on this PR) for argument parsing.

Sorry, I can not get your point, what is `clap routines` ?
I think it is a better way to parse argument when used.

jodh-intel: Well, I'm honestly rather confused. This PR:

- Adds a `src/enclave-agent/src/main.rs`, which defines a `main()` function.
- `main()` parses CLI arguments using `clap::App::new()`, and then starts the server.
- The server creates an instance of `ImageService` which defines a method called `pull_image()`.
- (_And now the bit I don't understand_): `pull_image()` *also* parses CLI options "manually":
  ```rust
  let args: Vec<String> = env::args().collect(); // get CLI args

  // Parse them
  let config_position = args.iter().position(|a| a == "--decrypt-config" || a == "-c");

  // ...
  ```

Hence, my questions are:

1) Why does `pull_image()` also need to parse CLI arguments? Why can't this all be handled in `main()` and passed to the `ImageService` service as a config struct or similar?
2) If you really, really need to do global CLI parsing in the server code, you can do it with clap. That would be consistent with `main()` and avoid "custom parsing" (why bother? we've got clap to help us! ;)
   ```rust
   async fn pull_image(&self, req: &image::PullImageRequest) -> Result<String> {
       let args: Vec<String> = env::args().collect();

       let matches = App::new("ocicrypt-config-parser").blah.blah.blah.get_matches_from(args);

       // ...
   }
   ```

YangLiang3: @jodh-intel  Hi Jiames, The config is not a global configuration, just in the pull image filed, so it's better to put the config file into the function filed rather than be global

haosanzi (approved):

src/libs/protocols/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+default = []
+with-serde = [ "serde", "serde_json" ]
+
+[dependencies]
+ttrpc = { git = "https://github.com/haosanzi/ttrpc-rust", features = ["async"] }
```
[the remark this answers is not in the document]
zhiwei-intel-h: > Could you specify a version for this forked repo? Also, is there an issue that you could add as a comment here for tracking the removal of this fork?

Ok，sure.  If the Occlum supports the Unix Socket, it will be replaced with official crate.

Issue #5, opened 2022-03-24
import rune with a clean base from IC project


jiazhang0:

jiazhang0:
This work is suspended. The first release of enclave-cc will use rune externally.

PR #16, opened 2022-07-15
Rune initial commit

YangLiang3:
1. add rune initial commit
2. change the path of go module reference
3. remove EPM(enclave pool management) function


jiazhang0:
Please add the "Fix" line to associate the issue https://github.com/confidential-containers/enclave-cc/issues/5

YangLiang3:
> Please add the "Fix" line to associate the issue #5 @YangLiang3

ok Got it

mythi:
To simplify the installation procedure and maintenance of `enclave-cc` I'd propose we go with `runc`. If `rune` is required at a later point, it might be better to use it from Inclavare containers two avoid the need for having to main two versions of `rune`.

YangLiang3:
> To simplify the installation procedure and maintenance of `enclave-cc` I'd propose we go with `runc`. If `rune` is required at a later point, it might be better to use it from Inclavare containers two avoid the need for having to main two versions of `rune`.
ok, got it. I prefer to use the rune as a default container runtime in enclave-cc. Because of the unchanged's entry point and the same UX.

YangLiang3:
> To simplify the installation procedure and maintenance of `enclave-cc` I'd propose we go with `runc`. If `rune` is required at a later point, it might be better to use it from Inclavare containers two avoid the need for having to main two versions of `rune`.

ok, got it. I prefer to use the rune as a default container runtime in enclave-cc. Because of the unchanged's entry point and the same UX

mythi:
> > To simplify the installation procedure and maintenance of `enclave-cc` I'd propose we go with `runc`. If `rune` is required at a later point, it might be better to use it from Inclavare containers two avoid the need for having to main two versions of `rune`.
> 
> ok, got it. I prefer to use the rune as a default container runtime in enclave-cc. Because of the unchanged's entry point and the same UX

The entry point change is not user visible or the `shim-rune` modification did not work?

YangLiang3:
> 1. add rune initial commit
> 2. change the path of go module reference
> 3. remove EPM(enclave pool management) function

YangLiang3:
> > > To simplify the installation procedure and maintenance of `enclave-cc` I'd propose we go with `runc`. If `rune` is required at a later point, it might be better to use it from Inclavare containers two avoid the need for having to main two versions of `rune`.
> > 
> > 
> > ok, got it. I prefer to use the rune as a default container runtime in enclave-cc. Because of the unchanged's entry point and the same UX
> 
> The entry point change is not user visible or the `shim-rune` modification did not work?

yes shim-rune can make some workaroud to this. But it needs to spend sometime on this.

Issue #15, opened 2022-07-13
Use Kata Containers rust crate for container ID verification


jodh-intel:
## Background

As mentioned in https://github.com/confidential-containers/enclave-cc/pull/14#discussion_r913100448, that PR includes some code from the Kata Containers [agent](https://github.com/kata-containers/kata-containers/tree/main/src/agent) (added on https://github.com/kata-containers/kata-containers/pull/1521).

## Upcoming Kata changes

Hence, there are now two versions of this code. However, the original version has been moved into a separate rust crate in the Kata 3.x `runtime-rs` branch:

- https://github.com/kata-containers/kata-containers/blob/runtime-rs/src/libs/kata-sys-util/src/validate.rs#L17

The plan is to merge the `runtime-rs` branch into Kata's `main` branch soon.

## CoCo Plan

Once  the `runtime-rs` branch has been merged into Kata's `main` branch, we should make this repo consume the `kata-sys-util` crate as a dependency.

## Why bother?

The function is only ~18 lines of code, so is it worth doing this? I would say yes for the following reasons:

### Unit tests

The original version of the code comes with a set of tests, whereas the version of the PR in this repo does not.

### Maintenance issues

If there are multiple copies of the code, who's going to maintain them and ensure they stay in sync, with the latest fixes and improvements?

Rust eschews the golang approach of code copying (aka "vendoring") since [crates.io](https://crates.io/) makes that approach unnecessary.

### Security

This is the more important reason. The backstory for my raising https://github.com/kata-containers/kata-containers/pull/1521 was that container / sandbox IDs have a habit of ending up as part of path names for a container/sandbox specific data store. In CLI parlance...

```bash
$ cid="foo"
$ tree "/run/coco/containers/${cid}/"
/run/coco/containers/foo/
├── bar.json
└── baz.config
```

At some point, that path will be deleted:

```bash
$ sudo rm -rf "/run/foo/containers/${cid}/"
```

That _seems_ reasonable. How about now?

```bash
$ cid='../../../sbin/init'
$ sudo /bin/rm "/run/foo/containers/${cid}/"
```

Ouch. Not so good. This is just an example, but the point is that the function in question provides a basic level of protection against these sorts of abuses (see the unit tests on the origional PR for further details).

/cc @YangLiang3, @quanweiZhou, @bergwolf for thoughts.

In summary, we should try to reuse code as much as possible in CoCo, but avoid copying it if at all possible.

[...]

PR #17, opened 2022-07-21
shim-rune: add init implementation

haosanzi:


src/shim/Makefile
```diff
@@ -0,0 +1,71 @@
+# Root directory of the project (absolute path).
+ROOTDIR=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+
+# Base path used to install.
+DESTDIR ?= /usr/local
+
+ifneq ($(GO111MODULE),off)
+	MOD_VENDOR := "-mod=vendor"
```
mythi, line 8: do we want/need to support vendored builds and keep `vendor/`?

src/shim/runtime/v2/rune/v2/service.go
```diff
@@ -447,6 +448,43 @@ func (s *service) Start(ctx context.Context, r *taskAPI.StartRequest) (*taskAPI.
 			Pid:         uint32(p.Pid()),
 		})
 	}
+
+	if strings.EqualFold(s.puaseID, r.ID) {
```
mythi, line 497: some consistent typo somewhere with: `s.puaseID`?

src/shim/runtime/v2/rune/config/config_test.go
```diff
@@ -0,0 +1,40 @@
+package config
+
+import (
+	"io/ioutil"
```
mythi, line 4: new code should not be using `io/ioutil` anymore.

mythi (commented):
@haosanzi thanks! I did a very quick review and opened some topics to discuss.

src/shim/runtime/v2/rune/v2/create.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, err
+	}
+
+	container, err := runc.NewContainer(ctx, s.platform, r)
+	if err != nil {
+		return nil, err
+	}
+
+	switch containerType {
+	case vc.PodSandbox:
```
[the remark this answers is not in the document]
haosanzi: Hi, Mikko. Thank you for your advice. Could you please explain in detail how to use `cri-api`? 
This container type is referred to [kata runtime](https://github.com/confidential-containers/kata-containers-CCv0/blob/CCv0/src/runtime/pkg/containerd-shim-v2/create.go#L79).

src/shim/Makefile
haosanzi: Hi, Mikko. The vendor mechanism solves the problem of reproducible construction of Go programs to a certain extent, however, its experience is not so good. The huge vendor directory needs to be submitted to the repo. Do you have any good suggestions, please? Thank you very much.

src/shim/runtime/v2/rune/v2/create.go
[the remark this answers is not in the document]
mythi: Looks like CRI's `type PodSandbox` is not what we want here, sorry. But I think this can be simplified so that there's no Kata dependency: looking at how Kata does it, it looks for a certain annotation in the OCI spec to check `config.json`is for a `PodSandbox` so that could be used directly here, right?

I noticed we have `loadSpec()` and `LoadSpec()` for doing the same thing.

src/shim/Makefile
mythi: My preference would be to not keep `vendor/` at least in the initial stage but I'm not aware of CoCo project needs to have support for offline builds etc. which might require it.

src/shim/runtime/v2/rune/v2/create.go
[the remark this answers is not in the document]
haosanzi: Thank you. Yes, In kata, it looks for a `"io.kubernetes.cri.container-type"` annotation in the OCI spec to check `config.json` is for a `PodSandbox`.   

Regarding whether to have Kata dependency, I have noticed [this issue](https://github.com/confidential-containers/enclave-cc/issues/15) about enclave-agent, and I agree that we should try to reuse code as much as possible in CoCo, but avoid copying it if at all possible.

image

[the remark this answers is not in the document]
mythi: My thinking was that there's no need to reuse nor copy if we just check:

```suggestion
switch ociSpec.Annotations["io.kubernetes.cri.container-type"] {
	case "PodSandbox":
	...
	case "PodContainer":
	...
}
```

do you think that would work?

src/shim/Makefile
YangLiang3: @mythi  still have some comments to this PR?

mythi: @YangLiang3 sorry I have not seen any updates being pushed. I will do the review shortly

src/shim/runtime/v2/rune/v2/create.go
```diff
@@ -2,16 +2,21 @@ package v2
...
 	"github.com/confidential-containers/enclave-cc/src/shim/runtime/v2/rune/config"
 	"github.com/confidential-containers/enclave-cc/src/shim/runtime/v2/rune/oci"
 	shimtypes "github.com/confidential-containers/enclave-cc/src/shim/runtime/v2/rune/types"
+	"github.com/containerd/containerd/api/types"
 	"github.com/containerd/containerd/runtime/v2/runc"
 	taskAPI "github.com/containerd/containerd/runtime/v2/task"
+	"github.com/sirupsen/logrus"
 )
 
+const ContainerBase = "/run/enclave-cc/app"
```
mythi, line 18: it must be possible to mock this dir so that the unit tests can run:
```
    create_test.go:313: 
        	Error Trace:	create_test.go:313
        	Error:      	"mkdir /run/enclave-cc: permission denied" does not contain "failed to mount rootfs component"
        	Test:       	TestCreateApp
    create_test.go:315: 
```

src/shim/runtime/v2/rune/constants/constants.go
```diff
@@ -0,0 +1,5 @@
+package constants
+
+const (
+	RuneOCIRuntime = "rune"
```
mythi: we have talked whether we could just go with `runc`. how about this is made configurable in this rev? it would be easier to, e.g., run the unit tests.

src/shim/Makefile
```diff
@@ -0,0 +1,70 @@
+# Root directory of the project (absolute path).
+ROOTDIR=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
```
jodh-intel, line 2: Nit: `ROOT_DIR` is easier to read.

src/shim/Makefile
```diff
@@ -0,0 +1,70 @@
...
+			endif
+			ifeq ($(UNAME_S),FreeBSD)
+				GOOS = freebsd
+			endif
+		endif
+	else
+		GOOS ?= $$GOOS
+		GOARCH ?= $$GOARCH
+	endif
+endif
```
jodh-intel, line 35: All these lines have a tab prefix which is not recommended for makefile variables.

src/shim/Makefile
```diff
@@ -0,0 +1,70 @@
...
+			ifeq ($(UNAME_S),Darwin)
+				GOOS = darwin
+			endif
+			ifeq ($(UNAME_S),FreeBSD)
+				GOOS = freebsd
+			endif
+		endif
+	else
+		GOOS ?= $$GOOS
+		GOARCH ?= $$GOARCH
```
jodh-intel, line 33: I don't think these lines are necessary? For example, if `GOOS` is already set, it's already accessible to `make` ;)

src/shim/Makefile
```diff
@@ -0,0 +1,70 @@
...
+
+.PHONY: clean all binaries help install uninstall package
+.DEFAULT: default
+
+all: binaries
+
+# Build a binary from a cmd.
+bin/containerd-shim-rune-v2:
+	@echo "bin/containerd-shim-rune-v2"
+	@CGO_ENABLED=${SHIM_CGO_ENABLED} GOOS=${GOOS} go build $(MOD_VENDOR) ${GO_BUILD_FLAGS} -o bin/containerd-shim-rune-v2 ./cmd/containerd-shim-rune-v2
```
jodh-intel, line 53: - Consistency: Please use either brackets or braces with variables, but consistently.
- I'd consider doing something like this:
  ```
  SHIM = containerd-shim-rune-v2
  COMMANDS += $(SHIM)
  ```
  Then,

```suggestion
	@CGO_ENABLED=${SHIM_CGO_ENABLED} GOOS=${GOOS} go build $(MOD_VENDOR) ${GO_BUILD_FLAGS} -o bin/$(SHIM) ./cmd/$(SHIM)
```

src/shim/Makefile
```diff
@@ -0,0 +1,70 @@
...
+
+binaries: clean $(BINARIES) ## build binaries
+
+clean: ## clean up binaries
+	@rm -f $(BINARIES)
+
+install: ## install binaries
+	@echo "$@ $(BINARIES)"
+	@mkdir -p $(DESTDIR)/bin
+	@install $(BINARIES) $(DESTDIR)/bin
```
jodh-intel, line 63: I'd consider making the installation explicit by specifying owner, group and full perms here.

src/shim/Makefile
```diff
@@ -0,0 +1,70 @@
...
+
+uninstall:
+	@echo "$@"
+	@rm -f $(addprefix $(DESTDIR)/bin/,$(notdir $(BINARIES)))
+
+help: ## this help
+	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST) | sort
```
jodh-intel, line 70: IOMCC? :smile:

src/shim/README.md
```diff
@@ -0,0 +1,34 @@
...
+## How to build and install
+
+### Step 1: Build and install shim binary.
+
+```bash
+git clone https://github.com/confidential-containers/enclave-cc
+
+cd src/shim
+GOOS=linux make binaries
+make install
```
jodh-intel, line 22: - This is an emotive topic, but I do wonder if we should use the Kata approach of always specifying a `$` prefix for commands the user should type:

  - https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md#code-blocks

  That makes it clear which lines are commands and which are command output, but does mean of course that you can't simply copy/paste the block into a terminal (although there are lots of standard JS widgets to toggle the prefix, but sadly GH doesn't offer that feature still).
- The `make install` will fail as written so I'd consider making that `sudo make install`.

src/shim/conf/config.toml
```diff
@@ -0,0 +1,10 @@
+log_level = "info" # "debug" "info" "warn" "error"
+
+[containerd]
+    # Pre-installed OCI bundle for agent enclave container
+    agent_container_instance = "/opt/enclave-cc/agent-instance/"
+    # The root dir of agent enclave container running state
+    agent_container_root_dir = "/run/containerd/agentenclave"
+    agent_url = "tcp://0.0.0.0:7788"
```
jodh-intel, line 8: Security: Does this need a comment explaining that the default URL will listen on all network interfaces?

src/shim/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+### Step 2: Configuration
+
+Modify containerd configuration file(/etc/containerd/config.toml) and add runtimes rune into it.
```
jodh-intel, line 27:
```suggestion
Modify the containerd configuration file (`/etc/containerd/config.toml`) and add the runtime rune to it.
```

src/shim/conf/config.toml
```diff
@@ -0,0 +1,10 @@
+log_level = "info" # "debug" "info" "warn" "error"
+
+[containerd]
+    # Pre-installed OCI bundle for agent enclave container
+    agent_container_instance = "/opt/enclave-cc/agent-instance/"
+    # The root dir of agent enclave container running state
+    agent_container_root_dir = "/run/containerd/agentenclave"
```
jodh-intel, line 7: Consistency + legibility: If this is a path we control (?), please add a dash in this to make it clearer: `agent-enclave`.

src/shim/conf/config.toml
```diff
@@ -0,0 +1,10 @@
+log_level = "info" # "debug" "info" "warn" "error"
```
jodh-intel, line 1: No "trace" and "critical" levels?

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Create dir for store unionfs image (based on sefs)
+	sefsDir := filepath.Join(c.Bundle, "rootfs/images/", cid, "sefs")
+	lowerDir := filepath.Join(sefsDir, "lower")
+	upperDir := filepath.Join(sefsDir, "upper")
+	for _, dir := range []string{lowerDir, upperDir} {
+		if err := os.MkdirAll(dir, 0711); err != nil && !os.IsExist(err) {
```
jodh-intel, line 162: That's a strange permission. I'd consider making that a global const and adding a comment. (Could it be `0700` btw?)

src/shim/runtime/v2/rune/v2/create.go
```diff
@@ -0,0 +1,140 @@
...
+			r.Rootfs = append(r.Rootfs, &types.Mount{
+				Type:    "overlay",
+				Source:  "overlay",
+				Options: options,
+			})
+
+			// Update the Root.Path field in container spec from
+			// "/var/lib/containerd/io.containerd.grpc.v1.cri/containers/<id>/rootfs" to "rootfs"
+			// TODO: config.json will be updated by agent enclave container
+			err = config.UpdateRootPathConfig(filepath.Join(r.Bundle, "config.json"), "rootfs")
```
jodh-intel, line 108: Please create a global const for `config.json` which is specified multiple times.

src/shim/runtime/v2/rune/v2/create.go
```diff
@@ -0,0 +1,140 @@
...
+			return nil, err
+		}
+	}
+
+	return container, nil
+}
+
+// ReadAgentIdFile reads the agent container id information from the path
+func readAgentIdFile(path string) (string, error) {
+	data, err := os.ReadFile(filepath.Join(path, "agent.id"))
```
jodh-intel, line 130: - What create this file?
- ooi, why is it not named something like `agent-id`?

src/shim/runtime/v2/rune/v2/create.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return "", err
+	}
+	return string(data), nil
+}
+
+// WriteAgentIdFile writes the agent container id information into the path
+func writeAgentIdFile(path, id string) error {
+	return os.WriteFile(filepath.Join(path, "agent.id"), []byte(id), 0600)
```
jodh-intel, line 139: Repeated hard-coded string. Please create a global const for `agent.id`.

jodh-intel (commented):
Thanks @haosanzi - a few comments...

src/shim/Makefile
YangLiang3: appreciate Mikko：）

src/shim/runtime/v2/rune/constants/constants.go
YangLiang3: hi Mikko
We will use rune in the first release. And runc will be the next plan. So let us use the rune at present

mythi: I think it needs to be at least configurable to get the unit tests in CI working without dependencies to rune.

jodh-intel:
Hi @haosanzi - any update on this?

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -18,9 +25,160 @@ import (
...
+	if cancel != nil {
+		defer cancel()
+	}
+	c.Logger().WithField("name", msgName).WithField("req", message.String()).Trace("sending request")
+
+	return handler(ctx, request)
+}
+
+func (c *agent) PullImage(ctx context.Context, req *image.PullImageReq) (*image.PullImageResp, error) {
+	cid, err := getContainerID(req.Image)
```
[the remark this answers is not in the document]
haosanzi: Shim will send the `PullImageReq` will sent to agent enclave container. 
The enclave agent will detect that the image has been pulled, then return PullImageResp response to the shim.

src/shim/runtime/v2/rune/v2/utils.go
```diff
@@ -46,3 +49,15 @@ func generateID() string {
 
 	return hex.EncodeToString(b)
 }
+
+func getContainerID(image string) (string, error) {
+	v := strings.Split(image, "/")
+	if len(v[len(v)-1]) <= 0 {
```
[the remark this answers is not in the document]
haosanzi: Hi, about the generate of `cid`, I referred to the implement in [kata-agent](https://github.com/confidential-containers/kata-containers-CCv0/blob/CCv0/src/agent/src/image_rpc.rs#L198). I think it is better to keep consistent. Thank you.

src/shim/Makefile
haosanzi: Hi, I refer to the [containerd](https://github.com/containerd/containerd/blob/main/Makefile#L459) implement, I don't understand the meaning of IOMCC very well. Could you please clarify it? Thank you very much.

src/shim/runtime/v2/rune/v2/create.go
haosanzi: When the agent enclave container is created (RunSanbox 
stage), the container id of agent enclave container id will be written in `agent-id` file (writeAgentIdFile function). 
According to the container in the `agent-id` file, we can get the resource of the agent enclave container, then clean up the resource of the agent container when stopping the pod sandbox.

haosanzi:
> Hi @haosanzi - any update on this?

Thank you for your comment, there was a very urgent project last week and I didn't have time to modify this pr in time. I'll have more time to revise the pr after this week. Thanks.

src/shim/runtime/v2/rune/v2/create.go
haosanzi: hi, does `mock this dir` means to create the dir manually or else? Could you please `sudo go test` to run the unit tests?

mythi: > hi, does `mock this dir` means to create the dir manually or else? Could you please `sudo go test` to run the unit tests?

for unit tests, I'm expecting them to create temporary directories for the tests to run and then have them to cleanup afterwards. Even if `sudo go test` worked, it's not a good idea for the tests to create some system directory and leave it there. Therefore, I was asking to change this so that for the tests, we have `ContainerBase` in some `<tmpdir>/run/enclave-cc/app`.

src/shim/runtime/v2/rune/v2/utils.go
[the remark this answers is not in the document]
mythi: @haosanzi my comment was more about how `v` is used multiple times to get the last element (image name).

src/shim/runtime/v2/rune/v2/service.go
```diff
@@ -47,6 +47,12 @@ var (
 	empty = &ptypes.Empty{}
 )
 
+// shimLog is logger for shim package
```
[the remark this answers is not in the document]
mythi: @haosanzi any thoughts on this? (I can see it was resolved but no changes to address this).

YangLiang3 (approved):

src/shim/runtime/v2/rune/v2/utils.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+
+	"github.com/BurntSushi/toml"
```
mythi, line 12: this triggers a build error

src/shim/runtime/v2/rune/v2/create_test.go
```diff
@@ -0,0 +1,141 @@
...
+	assert.NoError(err)
+
+	ociConfigFile = filepath.Join(bundlePath, "config.json")
+	err = os.WriteFile(ociConfigFile, busyboxConfigJson, testFileMode)
+	assert.NoError(err)
+
+	return tmpdir, bundlePath, ociConfigFile
+}
+
+func TestCreatePod(t *testing.T) {
```
mythi, line 60: I think we'll need to specify built-in defaults and don't fail on missing files:
```
    create_test.go:95: 
        	Error Trace:	create_test.go:95
        	Error:      	"open /etc/enclave-cc/config.toml: no such file or directory" does not contain "OCI runtime create failed"
        	Test:       	TestCreatePod
```

src/shim/runtime/v2/rune/v2/create_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	ctx := namespaces.WithNamespace(context.Background(), "UnitTest")
+	stubs := gostub.Stub(&ContainerBase, t.TempDir())
+	defer stubs.Reset()
+	_, err = s.Create(ctx, req)
+
+	expectedErr := "failed to mount rootfs component"
+	assert.Error(err)
+	assert.Contains(err.Error(), expectedErr)
```
mythi, line 137: use `FailNow()`if these fail since the assertions below are not relevant anymore

src/shim/runtime/v2/rune/v2/create_test.go
haosanzi: hi. In shim code, it will detect whether pre-installed OCI bundles (agent enclave instance and boot instance) and shim configure existed. If not, it will print the error. 

So if the developer and user want to test shim, they need to have pre-installed OCI bundles (agent enclave instance and boot instance) and shim configure first.
```
--- FAIL: TestCreatePod (0.00s)
    create_test.go:95: 
                Error Trace:    create_test.go:95
                Error:          "pre-installed OCI bundle should provided in /opt/enclave-cc/agent-instance/" does not contain "OCI runtime create failed"
```

mythi: > So if the developer and user want to test shim, they need to have pre-installed OCI bundles

I'm expecting that we setup our Github Actions CI to run these unit tests for every subsequent PR in this repository. That is why we need to make these tests to just run `go test -v ./...` without any dependencies to what is pre-installed. This case is similar to what we had with `/run/enclave-cc/`:  let's just have `<tmpdir>/opt/enclave-cc/agent-instance/` etc. created before running the tests.

src/shim/runtime/v2/rune/agent/client/client.go
```diff
@@ -0,0 +1,187 @@
...
+
+func parse(sock string) (string, *url.URL, error) {
+	addr, err := url.Parse(sock)
+	if err != nil {
+		return "", nil, err
+	}
+
+	var grpcAddr string
+	// validate more
+	switch addr.Scheme {
```
mythi, line 89: Supporting both `unix` and `tcp` looks like a lot of complexity especially because: 1) enclave-agent currently supports `tcp` only and 2) the mock server implemented in this PR only supports `unix`. Can we go with `tcp` (and have unit tests for it) and then do the switch.

src/shim/Makefile
jodh-intel: Sorry @haosanzi - it was a joke referring to https://en.wikipedia.org/wiki/International_Obfuscated_C_Code_Contest as this line is a bit cryptic :smile:

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Create dir for store unionfs image (based on sefs)
+	sefsDir := filepath.Join(c.Bundle, "rootfs/images/", cid, "sefs")
+	lowerDir := filepath.Join(sefsDir, "lower")
+	upperDir := filepath.Join(sefsDir, "upper")
+	for _, dir := range []string{lowerDir, upperDir} {
+		if err := os.MkdirAll(dir, 0700); err != nil && !os.IsExist(err) {
```
jodh-intel, line 164: I think the `os.IsExist(err)` check is redundant as `MkdirAll()` doesn't fail if the directory already exists. However, that _could_ be a problem if it exists and the permissions are not `0700` as we would not realise this.

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+	}
+
+	r := &grpc.PullImageRequest{
+		Image:       req.Image,
+		ContainerId: cid,
+	}
+	resp, err := c.sendReq(ctx, r)
+	if err != nil {
+		c.Logger().Errorf("agent enclave container pull image err. %v", err)
```
jodh-intel, line 175:
```suggestion
		c.Logger().WithError(err).Error("agent enclave container pull image)
```

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// The function creates agent enclave container based on a pre-installed OCI bundle
+func createAgentContainer(ctx context.Context, s *service, r *taskAPI.CreateTaskRequest) (*runc.Container, error) {
+	dir := filepath.Join(agentContainerRootDir, r.ID)
+	upperDir := path.Join(dir, "upper")
+	workDir := path.Join(dir, "work")
+	destDir := path.Join(dir, "merged")
+	for _, dir := range []string{upperDir, workDir, destDir} {
+		if err := os.MkdirAll(dir, 0700); err != nil && !os.IsExist(err) {
```
jodh-intel, line 191: There are multiple instances of `0700` on this PR so I'd consider creating a global `const` for it.

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -0,0 +1,279 @@
...
+		Source:  "overlay",
+		Options: options,
+	})
+	r.Bundle = destDir
+
+	fs.CopyFile(filepath.Join(r.Bundle, configFilename), filepath.Join(agentContainerPath, configFilename))
+
+	// Create Stdout and Stderr file for agent enclave container
+	r.Stdout = filepath.Join(agentContainerRootDir, r.ID, r.ID+"-stdout")
+	r.Stderr = filepath.Join(agentContainerRootDir, r.ID, r.ID+"-stderr")
```
jodh-intel, line 215: An ID is quite long so since we already have a host unique container directory, can we shorten these paths?

```suggestion
	r.Stdout = filepath.Join(agentContainerRootDir, r.ID, "stdout")
	r.Stderr = filepath.Join(agentContainerRootDir, r.ID, "stderr")
```

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -0,0 +1,279 @@
...
+		return err
+	}
+	opts, err := runc.ReadOptions(path)
+	if err != nil {
+		return err
+	}
+	root := process.RuncRoot
+	if opts != nil && opts.Root != "" {
+		root = opts.Root
+	}
```
jodh-intel, line 256: Nit: This code is a bit "dense". I'd consider a blank line between logical blocks to make it more readable.

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -0,0 +1,279 @@
...
+		"path":    path,
+		"ns":      ns,
+		"runtime": runtime,
+	}).Debug("agent enclave Container Cleanup()")
+
+	r := process.NewRunc(root, path, ns, runtime, "", false)
+	if err := r.Delete(ctx, id, &runcC.DeleteOpts{
+		Force: true,
+	}); err != nil {
+		logrus.WithError(err).Warn("failed to remove runc agent enclave container")
```
jodh-intel, line 269: Consistency: This is the only error message that mentions "runc".

src/shim/runtime/v2/rune/v2/create.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	sandboxNamspace, err := oci.SandboxNamespace(*ociSpec)
+	if err != nil {
+		return nil, err
+	}
+
+	var container *runc.Container
+
+	switch containerType {
+	case shimtypes.PodSandbox:
```
jodh-intel, line 37: These `case` blocks are too big imho. Could you refactor to call `handlePodSandbox()`, `handlePodContainer()`, _etc_?

src/shim/runtime/v2/rune/v2/create.go
jodh-intel: This still needs to reference `configFilename` I think?

src/shim/runtime/v2/rune/v2/create_test.go
```diff
@@ -0,0 +1,144 @@
...
+func SetupOCIConfigFile(t *testing.T) (rootPath string, bundlePath, ociConfigFile string) {
+	assert := assert.New(t)
+
+	tmpdir := t.TempDir()
+
+	bundlePath = filepath.Join(tmpdir, "bundle")
+	err := os.MkdirAll(bundlePath, testDirMode)
+	assert.NoError(err)
+
+	ociConfigFile = filepath.Join(bundlePath, "config.json")
```
jodh-intel, line 53: Can this reference `configFilename`?

src/shim/runtime/v2/rune/v2/utils.go
```diff
@@ -0,0 +1,70 @@
...
+	logLevel = cfg.LogLevel
+	agentContainerPath = cfg.Containerd.AgentContainerInstance
+	agentContainerRootDir = cfg.Containerd.AgentContainerRootDir
+	AgentUrl = cfg.Containerd.AgentUrl
+	bootContainerPath = cfg.Containerd.BootContainerInstance
+
+	for _, dir := range []string{agentContainerPath, bootContainerPath} {
+		fi, err := os.Stat(dir)
+		if err != nil {
+			return fmt.Errorf("pre-installed OCI bundle should provided in %s", dir)
```
jodh-intel, line 42:
```suggestion
			return fmt.Errorf("pre-installed OCI bundle should exist in %q", dir)
```

src/shim/runtime/v2/rune/v2/utils.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		if !fi.IsDir() {
+			return fmt.Errorf("not a directory: %s", dir)
```
jodh-intel, line 45:
```suggestion
			return fmt.Errorf("not a directory: %q", dir)
```

jodh-intel (commented):
Thanks @haosanzi - a few comments...

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -0,0 +1,279 @@
...
+)
+
+var (
+	defaultRequestTimeout = 60 * time.Second
+)
+
+type agent struct {
+	// ID of the container
+	ID string
+	// Bundle path
```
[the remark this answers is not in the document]
haosanzi: Absolute path in "agentContainerRootDir" in shim config (/run/containerd/agent-enclave/ )

src/shim/config/config.toml
```diff
@@ -0,0 +1,17 @@
...
+[containerd]
+    # Pre-installed OCI bundle for agent enclave container
+    agent_container_instance = "/opt/enclave-cc/agent-instance/"
+    # The root dir of agent enclave container running state
+    agent_container_root_dir = "/run/containerd/agent-enclave/"
+    # The listening address of the agent container
+    # Note that the default URL will listen on all network interfaces.
+    # TODO: After occlum releases NGO at the end of 2022, the URL will
+    # switch to the Unix domain socket.
+    agent_url = "tcp://0.0.0.0:7788"
```
mythi, line 15: we want localhost don't we
```suggestion
    agent_url = "tcp://127.0.0.1:7788"
```

src/shim/runtime/v2/rune/agent/client/client.go
```diff
@@ -0,0 +1,166 @@
...
+	ImageServiceClient agentgrpc.ImageService
+	conn               *ttrpc.Client
+}
+
+// NewAgentClient creates a new agent gRPC client and handles tcp address.
+//
+// Supported sock address format is:
+//   - tcp://<ip>:<port>
+func NewAgentClient(ctx context.Context, sock string, timeout uint32) (*AgentClient, error) {
+	grpcAddr, parsedAddr, err := parse(sock)
```
mythi, line 46: Is there a need for `parse()` anymore? It looks it just changes `tcp://foobar` to `tcp:foobar` from which we strip `tcp:` prefix. we could just call `TcpDialer()` (but maybe rename it)  directly instead?

src/shim/runtime/v2/rune/v2/create.go
```diff
@@ -0,0 +1,161 @@
...
+			URL:    AgentUrl,
+		}
+
+	}
+
+	return container, nil
+}
+
+func handlePodContainer(ctx context.Context, s *service, r *taskAPI.CreateTaskRequest, sandboxNamspace string, ociSpec *specs.Spec) (*runc.Container, error) {
+	if sandboxNamspace != shimtypes.KubeSystem {
```
mythi, line 92:
```suggestion
	if sandboxNamespace != shimtypes.KubeSystem {
```

mythi (commented):
remaining comments about the dialer structuring and just one nit about `namspace` typo

src/shim/Makefile
```diff
@@ -0,0 +1,69 @@
...
+
+# Get ARCH.
+ifeq ($(ARCH),)
+    ifneq (,$(golang_version_raw))
+        override ARCH = $(shell $(GO) env GOARCH)
+    else
+        override ARCH = $(shell uname -m)
+    endif
+endif
+ifeq ($(ARCH),x86_64)
```
jodh-intel, line 28: Nit: Missing blank line between blocks.

src/shim/runtime/v2/rune/config/config.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	return spec, nil
+}
+
+func SaveSpec(cPath string, spec *specs.Spec) error {
+	data, err := json.Marshal(spec)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(cPath, data, 0644)
```
jodh-intel, line 28: Could these perms be `0640` or even `0600`? I'd also consider making it a global `const` to make it as visible as possible (for auditing, and potential re-use).

src/shim/runtime/v2/rune/types/types.go
```diff
@@ -0,0 +1,13 @@
...
+const (
+	// PodSandbox identifies an infra container that will be used to create the pod
+	PodSandbox = "sandbox"
+	// PodContainer identifies a container that should be associated with an existing pod
+	PodContainer = "container"
+
+	// Default identifies the default namespace used by objects
+	Default = "default"
+	// KubeSystem identifies the namespace used by the Kubernetes system objects
+	KubeSystem = "kube-system"
```
jodh-intel, line 12: The values might be clearer if they had a prefix or suffix (`DefaultNS` and `KubeSystemNS`, for example).

src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+	}
+
+	var options []string
+	// Set index=off when mount overlayfs
+	options = append(options, "index=off")
+	options = append(options,
+		fmt.Sprintf("lowerdir=%s", filepath.Join(agentContainerPath, "rootfs")),
+		fmt.Sprintf("workdir=%s", workDir),
+		fmt.Sprintf("upperdir=%s", upperDir),
```
jodh-intel, line 205: `[file]path.Join()` would be safer here.

src/shim/runtime/v2/rune/v2/create.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		// sefsDir store the unionfs images (based on sefs)
+		sefsDir := filepath.Join(agentContainerRootDir, s.agentID, "merged/rootfs/images", cid)
+
+		var options []string
+		// Set index=off when mount overlayfs
+		options = append(options, "index=off")
+		options = append(options,
+			fmt.Sprintf("workdir=%s", workDir),
+			fmt.Sprintf("upperdir=%s", upperDir),
```
jodh-intel, line 118: Again, `[file]path.Join()` would be safer. It might be "impossible", but what if `r.ID == "foo %s bar"` for example. This function hasn't validated the request, so should trust it as little as possible.

src/shim/runtime/v2/rune/v2/create.go
```diff
@@ -0,0 +1,161 @@
...
+	data, err := os.ReadFile(filepath.Join(path, agentIDFile))
+	if err != nil {
+		return "", err
+	}
+	return string(data), nil
+}
+
+// WriteAgentIdFile writes the agent container id information into the path
+func writeAgentIdFile(path, id string) error {
+	return os.WriteFile(filepath.Join(path, agentIDFile), []byte(id), 0600)
```
jodh-intel, line 160: I'd consider making this perm a global `const`.

src/shim/runtime/v2/rune/v2/service.go
```diff
@@ -0,0 +1,1036 @@
...
+		ep, err = oomv1.New(publisher)
+	}
+	if err != nil {
+		return nil, err
+	}
+	go ep.Run(ctx)
+	s := &service{
+		id:         id,
+		context:    ctx,
+		events:     make(chan interface{}, 128),
```
jodh-intel, line 88: - Why 128?
- Could you make this a global `const` for clarity?

src/shim/runtime/v2/rune/v2/service.go
```diff
@@ -0,0 +1,1036 @@
...
+		return nil, err
+	}
+	args := []string{
+		"-namespace", ns,
+		"-id", id,
+		"-address", containerdAddress,
+	}
+	cmd := exec.Command(self, args...)
+	cmd.Dir = cwd
+	cmd.Env = append(os.Environ(), "GOMAXPROCS=4")
```
jodh-intel, line 154: - Why 4?
- Should this be configurable?
- Can you make this a global for visibility?

src/shim/runtime/v2/rune/v2/service.go
```diff
@@ -0,0 +1,1036 @@
...
+	}
+	defer f.Close()
+	var s spec
+	if err := json.NewDecoder(f).Decode(&s); err != nil {
+		return nil, err
+	}
+	return &s, nil
+}
+
+func (s *service) StartShim(ctx context.Context, opts shim.StartOpts) (_ string, retErr error) {
```
jodh-intel, line 174: This function is too big and complex imho. Could you split it up into more manageable pieces?

src/shim/runtime/v2/rune/v2/service.go
```diff
@@ -0,0 +1,1036 @@
...
+			}
+		}
+	}
+	if err := shim.AdjustOOMScore(cmd.Process.Pid); err != nil {
+		return "", fmt.Errorf("failed to adjust OOM score for shim: %w", err)
+	}
+	return address, nil
+}
+
+func (s *service) Cleanup(ctx context.Context) (*taskAPI.DeleteResponse, error) {
```
jodh-intel, line 288: Another very big function.

src/shim/runtime/v2/rune/v2/service.go
```diff
@@ -0,0 +1,1036 @@
...
+		Pid:        uint32(container.Pid()),
+	})
+
+	return &taskAPI.CreateTaskResponse{
+		Pid: uint32(container.Pid()),
+	}, nil
+}
+
+// Start a process
+func (s *service) Start(ctx context.Context, r *taskAPI.StartRequest) (*taskAPI.StartResponse, error) {
```
jodh-intel, line 435: Can you split this into smaller parts?

src/shim/runtime/v2/rune/v2/service.go
```diff
@@ -0,0 +1,1036 @@
...
+	logrus.WithField("container", r.ID).Debug("Start() start")
+	defer logrus.WithField("container", r.ID).Debug("Start() end")
```
jodh-intel, line 437: All these paired log calls make me wonder if adding tracing might be better medium term?

jodh-intel (commented):
Thanks @haosanzi - a few comments...

One general comment is that a lot of the functions in `src/shim/runtime/v2/rune/v2/service.go` are very big and ideally would be refactored into smaller parts to improve legibility and testability.

haosanzi:
> Thanks @haosanzi - a few comments...
> 
> One general comment is that a lot of the functions in `src/shim/runtime/v2/rune/v2/service.go` are very big and ideally would be refactored into smaller parts to improve legibility and testability.

Hi, James. Thank you for reviewing enclave-cc shim very much.

Since shim-rune inherits the implementation of shim-runc, it adds support for agent enclave and encrypted app enclave.
So the [first commit](https://github.com/confidential-containers/enclave-cc/pull/17/commits/b58a3c603d1b73b779b1d42c1d44972c32a39094) of shim pr (src/shim/runtime/v2/rune/v2/service.go) is almost the implementation of [containerd shim runc](https://github.com/containerd/containerd/blob/main/runtime/v2/runc/task/service.go#L60) with little modification.  

I think we'd better be consistent with the containerd shim runc implementation, which is easier to maintain. Thank you very much!

src/shim/runtime/v2/rune/agent/client/client.go
haosanzi: Hi, Mikko. In order to support both TCP and UNIX socket, it is better to add parse().  Libos may not support unix socket in some situations, user can change to TCP socket quickly.

src/shim/runtime/v2/rune/config/config.go
haosanzi: This file is the config.json of the container, 0644 should be better.

src/shim/runtime/v2/rune/agent/client/client.go
mythi: this initial version comes with TCP only so I was proposing to simplify code for that.

mythi (approved):
thanks!

jodh-intel (approved):
Thanks @haosanzi.

PR #52, opened 2022-10-31
build: allow run enclave-cc in SGX simulation mode

mythi:

### DIFF
--- a/tools/packaging/build/agent-enclave-bundle/Dockerfile
+++ b/tools/packaging/build/agent-enclave-bundle/Dockerfile
@@ -13,6 +13,7 @@ RUN git clone --depth 1 https://github.com/confidential-containers/enclave-cc.gi
 FROM ubuntu:20.04
 RUN apt-get update && \
     env DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    make \
     wget \
     gnupg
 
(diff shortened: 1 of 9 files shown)


tools/packaging/build/agent-enclave-bundle/Dockerfile
```diff
@@ -13,28 +13,31 @@ RUN git clone --depth 1 https://github.com/confidential-containers/enclave-cc.gi
...
-    libsgx-enclave-common \
-    libsgx-urts \
-    libsgx-uae-service \
-    libsgx-quote-ex \
     libsgx-dcap-ql \
     libsgx-dcap-quote-verify \
     libsgx-dcap-default-qpl
 
+WORKDIR /opt
+ARG SGX_SDK_INSTALLER=sgx_linux_x64_sdk_2.17.101.1.bin
```
qzheng527, line 29: @mythi The reason to install sgx sdk here is to support simulatuon mode, right?

mythi: Correct, I'm moving to use the SDK installer to get the `_sim.so` packages. I'm planning to work on an optimization later that drops unnecessary SDK content.

> Also is it better to use ARG to set the sgx version in the path below as well?

I could maybe add another `ARG` for that. Ideally only one for the version should be enough but annoyingly the SDK filename and the actual release version mismatch (`2.17.101.1` vs `2.17.1`)

mythi: looking at it again, I'm not sure adding a variable for that URL is useful since `2.17.1` is used in one place only

qzheng527: Or just put all the version related URL part as one ARG, such as 
```
ARG SGX_SDK_URL=https://download.01.org/intel-sgx/sgx-linux/2.17.1/distro/ubuntu20.04-server/sgx_linux_x64_sdk_2.17.101.1.bin
```
Then use `basename` to get the installer file name.
```
RUN wget $SGX_SDK_URL \
    && chmod +x $(basename $SGX_SDK_URL)
...
```

mythi: makes sense, I'll update!

mythi: updated.

mythi:
@qzheng527 looks OK now and ready for merge? I have a follow-up PR coming after this.

qzheng527:
> @qzheng527 looks OK now and ready for merge? I have a follow-up PR coming after this.

LGTM

haosanzi (approved):

mythi:
@haosanzi would you be able to merge?

YangLiang3 (approved):

PR #55, opened 2022-11-03
tools: Fix typo in the IMAGE env var name

fidencio:
IMAGEE was a typo, IMAGE should be the correct name.


fidencio:
By the way, mea culpa here as I pushed the content that I'm fixing mistakenly to the enclave-cc repo whilst I thought I was pushing to mine. /o\

PR #56, opened 2022-11-03
Automate payload generation -- 2nd try

fidencio:
This PR adds the needed fixes for automating the payload generation, be it on every push or when tagging a release.

Please, see each commit message for more details on what's been done on each commit.

https://github.com/fidencio/enclave-cc/actions/runs/3387943466/jobs/5629325048 is the current run where I'm trying to exercise this patches.

### DIFF
--- a/tools/packaging/build/build_payload.sh
+++ b/tools/packaging/build/build_payload.sh
@@ -2,6 +2,7 @@
 set -e
 
 CI=${CI:-no}
+PUSH=${PUSH:-no}
 SGX_MODE=${SGX_MODE:-HW}
 if [ "${CI}" == "yes" ]; then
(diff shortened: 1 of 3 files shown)


fidencio:
And https://quay.io/repository/confidential-containers/runtime-payload-ci?tab=tags shows it's already working!

tools/packaging/build/build_payload.sh
```diff
@@ -2,6 +2,7 @@
 set -e
 
 CI=${CI:-no}
+PUSH=${PUSH-no}
```
dcmiddle, line 5: does that need a colon?

fidencio: Oh, it does! :-)

mythi (approved):

PR #57, opened 2022-11-04
tools: use install to copy the installer script

mythi:


fidencio:
@mythi, I'd add to the commit message that using `install -D` fixes a problem that was introduced as part of https://github.com/confidential-containers/enclave-cc/commit/f227db05086c316f6a5c273b9c0cd5e08d5f9cbb.

In this way we make it clear for whoever will revisit this in a few months from now (probably ourselves) why we had to do that. :-)

mythi:
@fidencio updated

fidencio (approved):
lgtm, thanks @mythi!

PR #59, opened 2022-11-07
enclave-agent: Bump image-rs to `v0.2.0`

fidencio:
As image-rs has tagged its `v0.2.0` release, let's make sure we're using it in what will become the `v0.2.0` release of Enclave CC.


mythi (approved):
thanks!

Issue #60, opened 2022-11-09
image pull errors


mythi:
CoCo quickstart documentation uses `bitnami/nginx:1.22.0` image as an example and I gave it a try. I'm seeing different image pull errors:

```
Failed to pull image "docker.io/bitnami/nginx:1.22.0": rpc error: code = Internal desc = failed to mount "unionfs" to "/run/enclave-cc/containers/nginx_1.22.0/rootfs", with error: EIO: I/O error
```

to debug this in more details, I ran enclave-agent (`sudo runc run 123`) from the bundle with `OCCLUM_LOG_LEVEL=debug` and used the "async-client" to debug. This time I'm getting different errors:

```
Green Thread 1 - pull_image -> Err(RpcStatus(code: INTERNAL message: "unpack destination \"/var/lib/image-rs/layers/sha256_3a52f76b4a6462386fe51fabf6cc829dbdef540dc64cdcc809f907bfb6c68195\" already exists")) ended: 5.798667018s
```

The latter blocks me from investigating the former error in details.

haosanzi:
I have met this problem too. 
I think it's still a problem that the agent enclave has insufficient support for pulling multi-layer images from occlum Libos. 

I have also try the `docker.io/redis:latest` image. The error is:
```
Green Thread 1 - pull_image -> Err(RpcStatus(code: INTERNAL message: "failed to mount \"unionfs\" to \"/run/enclave-cc/containers/redis_latest/rootfs\", with error: EIO: I/O error")) ended: 16.988494372s
```

When I try to pull redis again, the error is
```
Green Thread 1 - pull_image -> Err(RpcStatus(code: INTERNAL message: "unpack destination \"/var/lib/image-rs/layers/sha256_f2cd78d6f24cd252df46286ce2bc350dd58893be20a39202b22e7e60b5c763d2\" already exists")) ended: 4.419962025s
```

qzheng527:
@haosanzi Can you help capture Occlum log when pulling redis? I can triage from Occlum side.

mythi:
@qzheng527 my logs (last 100 lines):
```
[2022-11-10T05:57:23.702Z][DEBUG][T3][#2692704][··Munmap] munmap: addr: 0x7f40e2000000, size: 0x2ac9000
[...]

PR #61, opened 2022-11-18
fix: verify container id

HaokunX-intel:
The container id is passed from the shim, the agent is able to verify it and unable to modify it.

Signed-off-by: Xing, Haokun <haokun.xin@intel.com>

### DIFF
--- a/src/enclave-agent/src/services/images.rs
+++ b/src/enclave-agent/src/services/images.rs
@@ -4,7 +4,7 @@ use std::io::prelude::*;
 use std::path::Path;
 use std::sync::Arc;
 
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use async_trait::async_trait;
 use image_rs::config::ImageConfig;
 use image_rs::image::ImageClient;
@@ -36,19 +36,7 @@ impl ImageService {
 
     async fn pull_image(&self, req: &image::PullImageRequest) -> Result<String> {
         let image = req.get_image();
-        let mut cid = req.get_container_id().to_string();
-
-        if cid.is_empty() {
-            let v: Vec<&str> = image.rsplit('/').collect();
-            if !v[0].is_empty() {
-                // ':' have special meaning for umoci during upack
-                cid = v[0].replace(":", "_");
-            } else {
-                return Err(anyhow!("Invalid image name: {:?}", image));
-            }
-        } else {
-            validate::verify_id(&cid)?;
-        }
+        let cid = self.get_container_id(req)?;
 
         let keyprovider_config = Path::new("/etc").join("ocicrypt_keyprovider_native.conf");
         if !keyprovider_config.exists() {
@@ -111,6 +99,14 @@ impl ImageService {
 
         Ok(image.to_owned())
     }
+
+    fn get_container_id(&self, req: &image::PullImageRequest) -> Result<String> {
+        let cid = req.get_container_id().to_string() ;
+        // keep consistent with the kata container convention, more details
+        // are described in https://github.com/confidential-containers/enclave-cc/issues/15
+        validate::verify_id(&cid)?;
+        Ok(cid)
+    }
 }
 
 #[async_trait]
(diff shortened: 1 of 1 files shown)


mythi (changes requested):
the generated code is likely not related to the fix. also, pls organize the commits so that the fix is visible

HaokunX-intel:
Need to rebase on the refactor issue #64.

src/enclave-agent/src/services/images.rs
```diff
@@ -111,6 +99,29 @@ impl ImageService {
...
+                cid = v[0].to_string()
+            } else {
+                return Err(anyhow!("Invalid image name: {:?}", image));  
+            }
+        }
+    
+        // FIXME: verify_id is borrowed from kata container CCv0
+        // which use umoci to unpack image. And ':' have special 
+        // meaning for umoci during upack. We need new functions 
+        // to generate cids and verify them in the future.
```
mythi, line 119: `verify_id` is used from `kata-sys-util` and addressed part of #15 so what this borrowing mean here?

src/enclave-agent/src/services/images.rs
```diff
@@ -111,6 +99,29 @@ impl ImageService {
 
         Ok(image.to_owned())
     }
+
+    fn parse_container_id(&self, req: &image::PullImageRequest) -> Result<String> {
+        let mut cid = req.get_container_id().to_string() ;
```
mythi, line 104: is the shim sending invalid container id?

src/enclave-agent/src/services/images.rs
HaokunX-intel: The "borrow" means that the function is defined in an outside crate `kata-sys-util` and we can not modify it directly. But as the #15 said, it is an acceptable design, I will reorganize the comment.

src/enclave-agent/src/services/images.rs
HaokunX-intel: The shim in this repo will format the container id in [here](https://github.com/confidential-containers/enclave-cc/blob/c9ab4f08cfcd7afdb66c93956150e71de6225446/src/shim/runtime/v2/rune/v2/utils.go#L58).  The function `parse_container_id` makes the last check.

mythi: I'm confused what is the error this PR fixes. If the shim sends valid `cid`s I don't see how this changes the flow. It also makes me think could we just error on all invalid (and empty) `cid`s

HaokunX-intel: According to the code, the original design mimics the kata on CCv0. The kata shim sends the pulling request to kata agent [here](https://github.com/confidential-containers/kata-containers/blob/cfa3e1e9334dedeb3e28440e6b3c3d268fc2ea14/src/runtime/pkg/containerd-shim-v2/image.go#L72), which has no container id but the image name. Therefore, the kata agent need to create and format the container id [here](https://github.com/confidential-containers/kata-containers/blob/cfa3e1e9334dedeb3e28440e6b3c3d268fc2ea14/src/agent/src/image_rpc.rs#L191) if the image name exists and the container id is empty. Otherwise the kata agent validates the container id if the the container id is provided. I find the problem because I send an invalid `cid` from a dev tool. The dev tool need to be fixed.

If we want  to error on all invalid (and empty) `cid`s, we need to enforce the client provides an non-empty `cid`. 

If we follow the kata design,  the PR can be closed.  I think it is a better choice because the it make the agent more flexible.

mythi: @HaokunX-intel thanks, great feedback. 

> If we want to error on all invalid (and empty) cids, we need to enforce the client provides an non-empty cid.
> 
> If we follow the kata design, the PR can be closed. I think it is a better choice because the it make the agent more flexible.

Our needs might be different from kata (e.g., image paths are shared on the host). One of the questions is do we need unique container ids paths independent of what the image name is? Can the image pull response `image_ref` used as the host path identifier instead of `container_id`? /cc @haosanzi 

We seem to have the container ID generation logic implemented in two places (shim and agent). With the kata approach our shim could drop the ID generation part but the current image-rs occlum implementation expects someone has created the sefs paths... and given that constraint the logic in agent-enclave could be to just check valid and non-empty `cid`s.

Given this it sounds to me this PR can change to just validate the `cid` is provided and it's valid. And then the dev tool needs fixing... I'm fine with the kata approach too but then we need to think how the sefs paths are created.

mythi: @haosanzi thanks, I also came to the same conclusion about the timing.

> Given this it sounds to me this PR can change to just validate the `cid` is provided and it's valid. And then the dev tool needs fixing...

do you agree this is the right approach then?

haosanzi: yes, I agree with you. Thank you very mych

HaokunX-intel: Codes have be updated.

src/enclave-agent/src/services/images.rs
```diff
@@ -111,6 +99,22 @@ impl ImageService {
 
         Ok(image.to_owned())
     }
+
+    fn get_container_id(&self, req: &image::PullImageRequest) -> Result<String> {
+        let cid = req.get_container_id().to_string() ;
+        
+        if cid.is_empty() {
```
mythi, line 106: empty `cid` is also detected by `verify_id` so maybe this check is not needed?

HaokunX-intel: Sure.

mythi: the formatting of the test cases looks strange. maybe we want to add a case for an empty container id as well?

mythi (approved):

haosanzi (approved):

dcmiddle (approved):

HaokunX-intel:
Fix third problem in issue #60 
> image tag detection is unstable

Issue #51, opened 2022-10-27
agent need to set image-rs and attestation agent config in a proper place


hairongchen:
now the ENABLE_SECURITY_VALIDATE  env var is set in pull_image and image-rs client(with validate image flag) is initialized before calling pull_image so failed to catch the env var hence the signature verification is not enabled/disabled according to agent's config

the agent's config should be read in right after the agent start to later set the config for image-rs and attestation agent

tasks to do
- [x]  rats-tls build into agent bundle to support EAA KBC (https://github.com/confidential-containers/enclave-cc/pull/48)
- [x]  operator to set agent config during runtime install (https://github.com/confidential-containers/enclave-cc/pull/83 and https://github.com/confidential-containers/operator/pull/147)
- [x]  agent to read config file https://github.com/confidential-containers/enclave-cc/pull/70

hairongchen-intel:
looking at the current agent code, the agent configuration file parse should happen right after the agent is running and before doing anything else

jiazhang0:
I'm not sure why ENABLE_SECURITY_VALIDATE is configurable through ENV variable. I mean the behavior of ENABLE_SECURITY_VALIDATE should be mandatory, rather than optional or configurable, unless the image-rs is built as debug mode.

hairongchen:
there is discussion in https://github.com/confidential-containers/image-rs/pull/60 and https://github.com/kata-containers/kata-containers/issues/4888 talking about the default behavior of signature validation
the intended solution is by default the validation is enabled but open a config option in agent config(both kata-agent and enclave-agent) to let user be able to disable the validation after serious consideration of the consequences or simply to easy get started on CoCo(thinking of the image signing solution in CoCo is both complicated and not completed yet)

Also currently image-rs does not provide a configurable interface for the enable/disable of validation, hence the ENV var is used to pass the info

mythi:
> there is discussion in [confidential-containers/image-rs#60](https://github.com/confidential-containers/image-rs/pull/60) 

based on this I believe this issue can be closed. agree?

hairongchen:
according to discussion at https://github.com/confidential-containers/image-rs/pull/60#discussion_r1023516669, we'll refine the agent to meet kata-agent behavior, @HaokunX-intel is working on it now.

hairongchen:
> > there is discussion in [confidential-containers/image-rs#60](https://github.com/confidential-containers/image-rs/pull/60)
> 
[...]

PR #62, opened 2022-11-18
WIP: move AA config away from Occlum instance

mythi:
To support the conversation in https://github.com/confidential-containers/enclave-cc/issues/51#issuecomment-1319480903


mythi:
covered by #83

PR #70, opened 2022-12-07
fix: load ImageService config

HaokunX-intel:
Load configs at the agent launch stage instead of
before the image pulling.

Signed-off-by: Xing, Haokun <haokun.xin@intel.com>


hairongchen:
cosign now is implemented as a feature of image-rs, so need to add to the feature list as bellow

> features = ["occlum_feature", "**cosign**"]

HaokunX-intel:
> cosign now is implemented as a feature of image-rs, so need to add to the feature list as bellow
> 
> > features = ["occlum_feature", "**cosign**"]

This PR loads config early and enables signature verification, regardless of client types ( `grpc-based client` or `native client`) or signature types (`single sign` or `cosign`). 

And this PR depends on `tag = 0.2.0` of `image-rs`, which has no `cosign` feature. We may propose another issue to enable `cosign`.

src/enclave-agent/src/config.rs
```diff
@@ -0,0 +1,119 @@
...
+            path: String,
+            result_key_provider: Option<String>,
+            result_security_validate: Option<bool>,
+        }
+
+        let config_cases: Vec<ConfigCase> = vec![
+            ConfigCase {
+                path: "test_data/decrypt_config/decrypt_config_normal.conf".to_string(),
+                result_key_provider: Some(
+                    "provider:attestation-agent:eaa_kbc::127.0.0.1:1234".to_string(),
```
Xynnn007, line 75: Could this constant `"provider:attestation-agent:eaa_kbc::127.0.0.1:1234"` be defined as a `const` in this test mod? Same fow below

src/enclave-agent/src/config.rs
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+}
+"#;
+
+#[derive(Serialize, Deserialize, Default)]
+pub struct DecryptConfig {
+    pub key_provider: Option<String>,
+    #[serde(default = "default_security_validate")]
+    pub security_validate: Option<bool>,
```
Xynnn007, line 19: Another flag is the switch for `auth`. related link for `ImageConfig` https://github.com/confidential-containers/image-rs/blob/main/src/config.rs#L29. I'm afraid we also need to add one flag for that?

Xynnn007 (commented):
Overall, LGTM.

Also, I think a configuration file/struct is good for image-rs. Maybe we can consider bring this mechinism directly into image-rs in the future to support
* more flexible configurations for image-rs, like auth
* counterpart for kata-cc

What is your opinion upon this? @arronwy

src/enclave-agent/src/config.rs
HaokunX-intel: The flag for `auth` has been added.

Xynnn007: Great, sorry for neglecting

Xynnn007 (approved):
Thanks @HaokunX-intel  LGTM

src/enclave-agent/src/services/images.rs
```diff
@@ -136,42 +100,45 @@ mod test {
 
     #[test]
     fn test_get_container_id() {
-        struct ParseCase{
+        struct ParseCase {
```
mythi, line 103: let's move formatting fixes to another pr

src/enclave-agent/src/config.rs
```diff
@@ -0,0 +1,126 @@
...
+        std::env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", self.config_path.clone());
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_load_decrypt_config() {
```
mythi, line 70: I think this is not testing what we are interested in:  errors are correctly returned when file read or json parsing fails based on inputs provided.

src/enclave-agent/Cargo.toml
```diff
@@ -20,14 +20,15 @@ ctrlc = { version = "3.0", features = ["termination"] }
 tokio = { version = "1.14.0", features = ["full"] }
 async-trait = "0.1.42"
 
-image-rs = { git = "https://github.com/confidential-containers/image-rs", tag = "v0.2.0", features = ["occlum_feature"], default-features = false }
+image-rs = { git = "https://github.com/confidential-containers/image-rs", "rev" = "78a5114", features = ["occlum_feature", "cosign"], default-features = false }
```
mythi, line 23: is it needed in this PR?

mythi (changes requested):

src/enclave-agent/Cargo.toml
HaokunX-intel: > Another flag is the switch for `auth`. related link for `ImageConfig` https://github.com/confidential-containers/image-rs/blob/main/src/config.rs#L29. I'm afraid we also need to add one flag for that?

The `auth` mentioned by @Xynnn007 is introduced after `tag = "v0.2.0"`.

src/enclave-agent/src/config.rs
HaokunX-intel: The rust standard create provides `file read` APIs, it is out of the responsibilities to cover all file read failures.

The crate `serde-json` provides `json parsing` APIs. It will return different error info once reading invalid json files, it is hard to emulate all cases.

I think we at most add two cases, including reading a not existing file, reading a json files with invalid key, to represent file read failure and json parse failure.

src/enclave-agent/src/services/images.rs
HaokunX-intel: I think these formatting fixes are too few to deserve to propose another PR.

mythi: a separate commit in this PR is also fine.

src/enclave-agent/src/config.rs
mythi: > I think we at most add two cases, including reading a not existing file, reading a json files with invalid key, to represent file read failure and json parse failure.

makes sense

src/enclave-agent/Cargo.toml
mythi: these are not related to the fix in this PR and in fact now it's blocking getting this PR merged. I'd suggest we drop `image-rs` updates in this PR so we get it merged.

src/enclave-agent/src/services/images.rs
HaokunX-intel: It sounds good.

src/enclave-agent/Cargo.toml
Xynnn007: Agreed. We can open an issue to trace registry authentication feature and impl in another PR.

HaokunX-intel: Sure, I'll drop the `auth` this PR.

src/enclave-agent/src/services/images.rs
mythi: all of these `test_get_container_id`  formatting changes can go into 5215746a32878641ce213086de3c2fce5af050bc

src/enclave-agent/src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+            } else {
+                let c = result.unwrap();
+                assert_eq!(c.key_provider, case.result_key_provider);
+                assert_eq!(c.security_validate, case.result_security_validate);
+            }
+        }
+    }
+
+    #[test]
+    fn test_keyprovider_config_default() {
```
mythi, line 125: It fails because the test case is not allowed to write to `DEFAULT_KEYPROVIDER_CONFIG_PATH`:

```
---- config::test::test_keyprovider_config_default stdout ----
thread 'config::test::test_keyprovider_config_default' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', src/config.rs:126:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

src/enclave-agent/src/main.rs
```diff
@@ -45,7 +48,16 @@ async fn main() -> Result<()> {
         TCP_SOCK_ADDR
     };
 
-    let image_service = Box::new(ImageService::new()) as Box<dyn image_ttrpc::Image + Send + Sync>;
+    let dc = if let Some(file_path) = matches.value_of("decrypt-config") {
+        DecryptConfig::load_from_file(&file_path.to_string())?
```
mythi, line 52: this path should handle errors better

src/enclave-agent/src/config.rs
HaokunX-intel: The test case is passed in our env. The agent should run in the root privilege to write to `/etc`. If we want to place the config file in different place, please propose another issue to track the concern.

src/enclave-agent/src/services/images.rs
HaokunX-intel: Actually, this PR affects almost all files.  It is hard and unbeneficial to separate new codes and format fixing into two independent parts completely. Therefore the last commit contains a part of formatting changes, moving all of them seems not necessary.

src/enclave-agent/src/config.rs
mythi: > The agent should run in the root privilege to write to `/etc`

this is not about running the agent but running the running tests. `cargo test` should pass without root/sudo privileges.

HaokunX-intel: If you insist it, please change the [enclave-agent.yaml](https://github.com/confidential-containers/enclave-cc/blob/572a5715bc4941c56e0b4dc0b4dacd3e5a9c4c42/tools/packaging/build/agent-enclave-bundle/enclave-agent.yaml#L25) , copy the ` /run/enclave-agent/decrypt_config.conf` to other directory instead of `/etc`.

src/enclave-agent/src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+        let path = DEFAULT_KEYPROVIDER_CONFIG_PATH.to_string();
+        let mut kc = Self {
+            config_path: path.clone(),
+            is_exist: Path::new(&path).exists(),
+        };
+
+        if !kc.is_exist {
+            File::create(Path::new(&path))?.write_all(DEFAULT_KEYPROVIDER_CONFIG.as_bytes())?;
+            kc.is_exist = true;
+        }
```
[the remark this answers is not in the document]
mythi: I'm not sure this answers my question. There's the option of giving an error if there's no file pre-installed.

src/enclave-agent/src/services/images.rs
mythi: it's easy to pick the lines to you want using `git add -p ...`

src/enclave-agent/src/main.rs
HaokunX-intel: Sure. The error will be handled in the `load_from_file`, this path will raise the error if failed.

src/enclave-agent/src/services/images.rs
HaokunX-intel: Sure.

src/enclave-agent/src/config.rs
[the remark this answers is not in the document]
HaokunX-intel: > I'm not sure this answers my question. There's the option of giving an error if there's no file pre-installed.

That is a good idea. But, nowadays we can not customize the config for `ocicrypt-rs`. And we do not provide a default config in the [packaging](https://github.com/confidential-containers/enclave-cc/tree/main/tools/packaging/build).  If the writeback is removed, the agent can not pull and decrypt any encrypted image.

If we want to remove the writeback and pull the encrypted images, we need to add a default config file in the packaging. But if we plan to provide an option to customize the ocicrypt-rs config in the future, the writeback is acceptable.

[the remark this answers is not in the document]
mythi: We do provide a default config: https://github.com/confidential-containers/enclave-cc/blob/572a5715bc4941c56e0b4dc0b4dacd3e5a9c4c42/tools/packaging/build/agent-enclave-bundle/enclave-agent.yaml#L25-L28

#62 is also experimenting with the idea of moving the default away from the Occlum instance. Check the plans how this is going to be used in the operator from @hairongchen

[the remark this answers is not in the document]
mythi: Yeah, sorry I mixed the two. We can install the keyprovider config too.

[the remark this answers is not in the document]
HaokunX-intel: If the keyprovider config is installed by the operator, we can abandon the writeback.

And a new launch option for loading keyprovider config can be provided.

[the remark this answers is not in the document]
HaokunX-intel: In this PR, we can install the default keyprovider config in the agent bundle. And #62 supports customized agent config and keyprovider config.

src/enclave-agent/src/config.rs
HaokunX-intel: The keyprovider config will not be writeback, the unit test can be run without root privilege now.

src/enclave-agent/src/services/images.rs
```diff
@@ -173,7 +134,11 @@ mod test {
             },
         ];
 
-        let is = ImageService::new();
+        let dc = DecryptConfig::load_from_file(
+            &"test_data/decrypt_config/decrypt_config_normal.conf".to_string(),
```
mythi, line 138: This should be the user provided `decrypt-config` param instead, right?

HaokunX-intel: Yeah, the user provides the file and specify the path the file when the agent is launched.

mythi (approved):

Issue #72, opened 2022-12-12
enclave-agent build error.


jiere:
Run "make" under src/enclave-agent, met below errors:

error[E0425]: cannot find function `decode_config` in crate `base64`
   --> /home/jie/.cargo/registry/src/github.com-1ecc6299db9ec823/sequoia-openpgp-1.11.0/src/armor/base64_utils.rs:166:19
    |
166 |     match base64::decode_config(&bytes, base64::STANDARD) {
    |                   ^^^^^^^^^^^^^ not found in `base64`

error[E0425]: cannot find value `STANDARD` in crate `base64`
   --> /home/jie/.cargo/registry/src/github.com-1ecc6299db9ec823/sequoia-openpgp-1.11.0/src/armor/base64_utils.rs:166:49
    |
166 |     match base64::decode_config(&bytes, base64::STANDARD) {
    |                                                 ^^^^^^^^ not found in `base64`
    |
help: consider importing this constant
    |
1   | use base64::alphabet::STANDARD;
    |
help: if you import `STANDARD`, refer to it directly
    |
166 -     match base64::decode_config(&bytes, base64::STANDARD) {
166 +     match base64::decode_config(&bytes, STANDARD) {
    |

hairongchen:
this crate is used in image-rs 
@Xynnn007 @arronwy

hairongchen:
base info:
hrc@cre-M50CYP2SBSTD:/tmp/image-rs$  cargo tree |grep  sequoia-openpgp
├── sequoia-openpgp v1.11.0


hrc@cre-M50CYP2SBSTD:/tmp/image-rs$ cargo tree |grep base6
│   ├── base64 v0.13.1
│   │   │   │       └── base64ct v1.5.3
│   │   │   │   │   ├── base64ct v1.5.3
├── base64 v0.13.1
│   │   ├── base64 v0.13.1
│   │   ├── base64 v0.13.1
│   ├── base64 v0.13.1
│   ├── base64-serde v0.6.1
│   │   ├── base64 v0.13.1
│   │   ├── base64 v0.13.1
│       ├── base64 v0.13.1
│   ├── base64 v0.20.0
│   │   │   ├── base64 v0.13.1
│   ├── base64 v0.13.1
│   │   ├── base64 v0.13.1
│   │   │   ├── base64 v0.13.1
│   │   └── base64 v0.13.1
│   │   ├── base64 v0.13.1
│   │   │   ├── base64 v0.13.1
[...]

PR #69, opened 2022-12-07
Enhancement: build payload artifacts in container.

jiere:
Closes: #45 
Signed-off-by: Jie Ren <jie.ren@intel.com>

PR #74, opened 2022-12-12
fix: build error for dependency error of enclave-agent

Xynnn007:
Close https://github.com/confidential-containers/enclave-cc/issues/72

As best-practice for rust, a binary should include `Cargo.lock`, however a crate module should not.

Signed-off-by: Xynnn007 <xynnn@linux.alibaba.com>


mythi:
we depend on #69 to get this tested properly.

Xynnn007:
> we depend on #69 to get this tested properly.

Yes? Sorry I can not get your point. The issue related to this PR seems to be caused by a dependency configuration. :)

mythi:
> > we depend on #69 to get this tested properly.
> 
> Sorry, I get your point. The issue here is cause by a dependency modification **one day ago**, so that link you've given passed for it occured **five days ago**.

without #69 we cannot get CI all passing for this PR because right now CI builds enclave-agent from `main` and not from this PR.

Xynnn007:
> without #69 we cannot get CI all passing for this PR because right now CI builds enclave-agent from `main` and not from this PR.

Great, I just found this problem. Thanks so much!

mythi:
@Xynnn007 pls rebase to retrigger the build

Issue #76, opened 2022-12-13
Install the RATS-TLS library in compile env to fix dependency bugs.


HaokunX-intel:
Although the PR #48 installs RATS-TLS in the runtime env, if we want to build a enclave agent referring to latest image-rs, which depends on the RATS-TLS, we need to install the RATS-TLS in the compile env too.

Otherwise, the ld complains that it can not find `rats-tls` as shown in [here](https://github.com/confidential-containers/enclave-cc/actions/runs/3683353586/jobs/6231836441)
```
 = note: /usr/bin/ld: cannot find -lrats_tls
          collect2: error: ld returned 1 exit status
```

The building [Dockerfile](https://github.com/confidential-containers/enclave-cc/blob/main/tools/packaging/build/agent-enclave-bundle/Dockerfile) should install the rats-tls in the builder, it may look like
```
FROM rust:1.63-bullseye as builder


RUN apt-get update && \
    env DEBIAN_FRONTEND=noninteractive apt-get install -y \
    protobuf-compiler

# FIX: install rats-tls
RUN git clone --depth 1 https://github.com/inclavare-containers/rats-tls.git && \ 
    cd rats-tls && \
    cmake -DRATS_TLS_BUILD_MODE="occlum" -DBUILD_SAMPLES=on -H. -Bbuild && \
    make -C build install

# Build enclave-agent
COPY src/ /enclave-cc/src/
RUN cd /enclave-cc/src/enclave-agent && \
    rustup component add rustfmt && \
    make

# Start preparing enclave-agent "bundle"
FROM ubuntu:20.04
...

```

mythi:
Would it be possible to get `rats-tls` installable from their Deb repo? http://mirrors.openanolis.cn/inclavare-containers/deb-repo/pool/main/r/rats-tls/

haosanzi:
hi, rats-tls can provide deb package. Put it in the repo of ic. 

just FYI: rats-tls will be updated recently to support the DICE standard. The relevant pr is at https://github.com/inclavare-containers/rats-tls/pull/123

HaokunX-intel:
I agree with @haosanzi . My concern is updating official repo may bring more overhead, which is too heavy to us currently. Given that rats-tls and enclave-cc update rapidly, we need more flexible paths to combine them. I think we can build the rats-tls with certain commit id  from source code to avoid compatible problems.

[...]

PR #77, opened 2022-12-20
fix: refer new image-rs

HaokunX-intel:
The new rev of the image-rs introduces the rats-tls, to refer it, we install rats-tls in the complie env and update the reference to the image-rs.

Signed-off-by: Xin, Haokun <haokun.xin@intel.com>

Closes: #60

### DIFF
--- a/src/enclave-agent/Cargo.toml
+++ b/src/enclave-agent/Cargo.toml
@@ -20,7 +20,7 @@ ctrlc = { version = "3.0", features = ["termination"] }
 tokio = { version = "1.14.0", features = ["full"] }
 async-trait = "0.1.42"
 
-image-rs = { git = "https://github.com/confidential-containers/image-rs", tag = "v0.2.0", features = ["occlum_feature"], default-features = false }
+image-rs = { git = "https://github.com/confidential-containers/image-rs", features = ["occlum_feature"], default-features = false, rev = "cf1f7f96eb60ec4cc4aaa671c04d574a4ea0e441" }
 kata-sys-util = { git = "https://github.com/kata-containers/kata-containers", rev = "4b57c04c3379d6adc7f440d156f0e4c42ac157df" }
 
 anyhow = "1.0.32"
--- a/tools/packaging/build/agent-enclave-bundle/Dockerfile
+++ b/tools/packaging/build/agent-enclave-bundle/Dockerfile
@@ -1,5 +1,7 @@
 FROM ubuntu:20.04 as builder
 
+ARG RATS_TLS_REV=5de6fc308f2a18f4105429451c85e6173b42ab9d
+
 RUN apt-get update && \
     env DEBIAN_FRONTEND=noninteractive apt-get install -y \
     make \
@@ -46,8 +48,9 @@ RUN env DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommend
     occlum-toolchains-glibc \
     occlum
 
-RUN git clone --depth 1 https://github.com/inclavare-containers/rats-tls.git && \
+RUN git clone https://github.com/inclavare-containers/rats-tls.git && \
     cd rats-tls && \
+    git checkout $RATS_TLS_REV && \
     cmake -DRATS_TLS_BUILD_MODE="occlum" -DBUILD_SAMPLES=on -H. -Bbuild && \
     make -C build install
 
@@ -83,5 +86,4 @@ RUN apt-get purge -y wget gnupg tzdata jq occlum occlum-toolchains-glibc make bi
 RUN echo "/run/rune/occlum_instance/build/lib/" | tee /etc/ld.so.conf.d/occlum-pal.conf && \
     echo "/opt/sgxsdk/lib64" | tee /etc/ld.so.conf.d/sgxsdk.conf && \
     ldconfig
-
 ENTRYPOINT ["/bin/enclave-agent"]
--- a/src/enclave-agent/Cargo.lock
+++ b/src/enclave-agent/Cargo.lock
@@ -167,42 +167,19 @@ dependencies = [
 ]
 
 [[package]]
-name = "asn1-rs"
-version = "0.5.1"
+name = "async-compression"
+version = "0.3.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf6690c370453db30743b373a60ba498fc0d6d83b11f4abfd87a84a075db5dd4"
+checksum = "942c7cd7ae39e91bde4820d74132e9862e62c2f386c3aa90ccf55949f5bad63a"
 dependencies = [
- "asn1-rs-derive",
- "asn1-rs-impl",
- "displaydoc",
- "nom",
- "num-traits",
- "rusticata-macros",
- "thiserror",
- "time 0.3.17",
-]
-
-[[package]]
-name = "asn1-rs-derive"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "726535892e8eae7e70657b4c8ea93d26b8553afb1ce617caee529ef96d7dee6c"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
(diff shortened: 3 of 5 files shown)


tools/packaging/build/agent-enclave-bundle/Dockerfile
```diff
@@ -1,5 +1,12 @@
 FROM rust:1.63-bullseye as builder
 
+RUN git clone --depth 1 https://github.com/inclavare-containers/rats-tls.git && \
```
haosanzi, line 3: It is recommended to use an env to represent the commit id of rats-tls, which is more conducive to later maintenance (In line 5 and line 66)，such as:
```
ENV RATS_TLS_REV 5de6fc308f2a18f4105429451c85e6173b42ab9d
git checkout $RATS_TLS_REV (In line 5 and line 66)
```

tools/packaging/build/agent-enclave-bundle/Dockerfile
```diff
@@ -1,5 +1,12 @@
...
+    cd rats-tls && \
+    git checkout 5de6fc308f2a18f4105429451c85e6173b42ab9d && \
```
mythi, line 5: since we clone with `--depth 1` this commit stops working when new content is merged. For example, `8cfcf9bf9f7902f85c921824c420c3c380939b06` does not work anymore.

haosanzi (approved):

tools/packaging/build/agent-enclave-bundle/Dockerfile
dcmiddle: Why are we building at the commit ID where RA-TLS is updated? i.e., because that commit is in main why don't we just build at the current head commit? We may miss future bug fixes if we peg to that commit.

Also per @mythi 's earlier comment, if we do not clone the commit history we will not clone this commit.

dcmiddle: Disregard my question about building against a specific commit. I see now in https://github.com/confidential-containers/enclave-cc/issues/76 where @mythi recommended doing that to avoid breaking our build if main is volatile. We will treat this commit ID like a stable tag.

My other comment about shallow clone not getting this commit is still relevant, though.

HaokunX-intel: Sure, option `--depth 1` has been removed.

mythi: One thing I was not very clear in [my comment](https://github.com/confidential-containers/enclave-cc/issues/76#issuecomment-1350483311) was related to "let's not do it twice". I wanted to say that instead of installing rats-tls/SGXSDK twice, we should move away from `rust:1.63-bullseye` and just install rust ourselves.

mythi: > we should move away from `rust:1.63-bullseye` and just install rust ourselves.

I'll submit a PR about this.

mythi: see #80

src/enclave-agent/Cargo.toml
```diff
@@ -20,7 +20,7 @@ ctrlc = { version = "3.0", features = ["termination"] }
 tokio = { version = "1.14.0", features = ["full"] }
 async-trait = "0.1.42"
 
-image-rs = { git = "https://github.com/confidential-containers/image-rs", tag = "v0.2.0", features = ["occlum_feature"], default-features = false }
+image-rs = { git = "https://github.com/confidential-containers/image-rs", features = ["occlum_feature"], default-features = false }
```
mythi, line 23: we want to use a known `rev` here

dcmiddle (approved):

src/enclave-agent/Cargo.toml
HaokunX-intel: `rev` has been added.

mythi: thanks, let's merge #80 first and then rebase this on top of it

src/enclave-agent/Cargo.toml
```diff
@@ -20,7 +20,7 @@ ctrlc = { version = "3.0", features = ["termination"] }
 tokio = { version = "1.14.0", features = ["full"] }
 async-trait = "0.1.42"
 
-image-rs = { git = "https://github.com/confidential-containers/image-rs", tag = "v0.2.0", features = ["occlum_feature"], default-features = false }
+image-rs = { git = "https://github.com/confidential-containers/image-rs", features = ["occlum_feature"], default-features = false, rev = "771e19820ad04946b28b590c3371cf1d765bd14d" }
```
fidencio, line 23: For Kata Containers we've bumped the image-rs to cf1f7f96eb60ec4cc4aaa671c04d574a4ea0e441, as that brings in the image layer stream pulling support.

It may be worth doing the same right now.

This will require a bump on the rust toolchain version to the 1.66.0

mythi:
@HaokunX-intel this needs rebasing after #80

HaokunX-intel:
> @HaokunX-intel this needs rebasing after #80

Rebase has been completed.

tools/packaging/build/agent-enclave-bundle/Dockerfile
```diff
@@ -1,5 +1,9 @@
+ARG RATS_TLS_REV=5de6fc308f2a18f4105429451c85e6173b42ab9d
+
 FROM ubuntu:20.04 as builder
 
+ARG RATS_TLS_REV
```
mythi, line 5: maybe just define once:
```suggestion
ARG RATS_TLS_REV=5de6fc308f2a18f4105429451c85e6173b42ab9d
```

HaokunX-intel: The `RATS_TLS_REV`  is designed for the two-stage building. The `RATS_TLS_REV` in defined and initialized in line 1. And then it can be used directly in the builder and the real container by re-declaring it again. I think the design can be reserved.

mythi: what is the use-case where it is needed outside the builder stage?

HaokunX-intel: For example, the agent depends on the dynamic lib `rats-tls`, the lib need to be installed in the builder to complete compiling, and in the real container to enable the agent running upon the bundle. 

And It is duplicated to define and initialize to two `RATS_TLS_REV` twice in the builder and the real container respectively, which looks like:
```
From ... as builder
ARG RATS_TLS_REV=5de6fc308f2a18f4105429451c85e6173b42ab9d

From ...
ARG RATS_TLS_REV=5de6fc308f2a18f4105429451c85e6173b42ab9d
```
It is also easy for us to forget to  update these two `RATS_TLS_REV` at the same time.

But our design will looks like:
```
ARG RATS_TLS_REV=5de6fc308f2a18f4105429451c85e6173b42ab9d
From ... as builder
ARG RATS_TLS_REV

From ...
ARG RATS_TLS_REV
```
It remedies the downside of the former design.

mythi: > And It is duplicated to define and initialize to two `RATS_TLS_REV` twice in the builder and the real container

AFAICS it's only needed once (in the builder).

HaokunX-intel: But as I know, the lib `rats-tls` installed in the builder will not appear in the real container, when the agent is launched, the `ld` can not find the `rats-tls`.

mythi: That's not correct. It needs to be installed in the Occlum instance only and the Occlum package (after it's built) containers `rats-tls`  and can be copied anywhere.

HaokunX-intel: If the builder is responsible for compiling the agent and packaging the occlum instance, the `rats-tls` can be installed on once.

mythi (approved):

PR #80, opened 2023-01-04
build: make RATS-TLS available for enclave-agent build

mythi:

### DIFF
--- a/tools/packaging/build/agent-enclave-bundle/Dockerfile
+++ b/tools/packaging/build/agent-enclave-bundle/Dockerfile
@@ -1,22 +1,13 @@
-FROM rust:1.63-bullseye as builder
+FROM ubuntu:20.04 as builder
 
-# Build enclave-agent
-RUN apt-get update && \
-    env DEBIAN_FRONTEND=noninteractive apt-get install -y \
-    protobuf-compiler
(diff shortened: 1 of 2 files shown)


tools/packaging/build/agent-enclave-bundle/Dockerfile
```diff
@@ -1,22 +1,13 @@
-FROM rust:1.63-bullseye as builder
+FROM ubuntu:20.04 as builder
```
HaokunX-intel, line 1: Given that we make an one-stage building, `as builder` can be removed.

mythi: good comment. I think we could keep it because we may want to go back to a multi-stage build where in the second stage only the instance and runtime dependencies get installed.

HaokunX-intel: Sure.

fidencio (approved):
lgtm, @mythi!

One thing that we may want to do in the future, not as part of this PR, is pin a rust version being used.  Right now I see you're installing the latest stable (1.66.0), which can start giving us new warnings when the next stable is out.

dcmiddle (approved):

PR #82, opened 2023-01-06
Fix directory not found issue in install script.

jiere:
The "/opt/confidential-containers/bin" directory was missed to be created before installing binaries there.

Signed-off-by: Jie Ren <jie.ren@intel.com>


tools/packaging/deploy/enclave-cc-deploy.sh
```diff
@@ -57,6 +57,7 @@ function install_artifacts() {
 
 	cp shim-rune-config.toml /etc/enclave-cc/config.toml
 
+	mkdir -p /opt/confidential-containers/bin
 	install -D -m0755 ${shim_rune_binary} /opt/confidential-containers/bin/
```
fidencio, line 61: I actually would make sure we have this line as:
```
install -D -m0755 ${shim_rune_binary} /opt/confidential-containers/bin/${shim_rune_binary}
```

That will do the trick without the need of adding the `mkdir`, as ...
```
-D
    create all leading components of DEST except the last, then copy SOURCE to DEST
```

mythi: +1

mythi:
@jiere can you update this so we get it merged?

fidencio:
I've just updated @jiere's PR according to the review.

mythi (approved):

PR #90, opened 2023-01-12
fix sgx path and config PCCS

huliucheng1:
Signed-off-by: Liucheng Hu <liucheng.hu@intel.com>

Fix sgx path and config PCCS file.


hairongchen:
@mythi  @haosanzi

PR #91, opened 2023-01-12
ci: do not test with Go 1.16 anymore

mythi:
newer golang.org/x/sys depend on Go 1.17+ versions. Since 1.16 is no longer supported, we can just drop it in our CI to get golang.org/x/sys and other dependencies updated.


fidencio (approved):
lgtm, thanks @mythi!

PR #97, opened 2023-01-18
Implement a simple CI pipeline.

huliucheng1:
Signed-off-by: Liucheng Hu <liucheng.hu@intel.com>

This is a simple CI pipeline, including install/uninstall the operator, cc-runtime, and apply/delete a workload with encryption and cosign signed.

### DIFF
--- a/.github/workflows/test_ci_cases.yaml
+++ b/.github/workflows/test_ci_cases.yaml
@@ -1,0 +1,46 @@
+name: CI tests
+
+on: 
+  push:
+    branches: [ "main" ]
+
+  pull_request:
+    branches: [ "main" ]
+
+jobs: 
+  CI-tests:
+    runs-on: self-hosted
+    steps: 
+
+      - name: Checkout latest enclave-cc source code
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v3
+        with:
+          ref: main
+
+      - name: Checkout PR source code
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+                
+      - name: Install CoCo operator
+        run: ./ci/ci_tests.sh install_coco_operator
+
+      - name: Install Enclave-CC runtimeclass payload
+        run: ./ci/ci_tests.sh install_enclave_cc_runtimeclass
+
+      - name: Pull and run a workload image signed by cosign
+        run: ./ci/ci_tests.sh apply_eaa_cosign_encryped_hello_world_workload
+
+      - name: delete workload pod
+        run: |
+          ./ci/ci_tests.sh delete_eaa_cosign_encrypted_hello_world_workload
+          sleep 10
+        
+      - name: Uninstall Enclave-CC runtimeclass payload 
+        run: ./ci/ci_tests.sh uninstall_enclave_cc_runtimeclass
+
+      - name: Uninstall CoCo operator
+        run: ./ci/ci_tests.sh uninstall_coco_operator 
+     
--- a/ci/utils.sh
+++ b/ci/utils.sh
@@ -1,0 +1,123 @@
+#!/bin/bash
+
+is_pod_exist() {
+    local POD_NAME=$1
+    kubectl get pods -A 2>&1 | grep $POD_NAME
+    return $?
+}
+
+is_runtimeclass_exist() {
+    local RUNTIMECLASS_NAME=$1
+    kubectl get runtimeclass 2>&1 | grep $RUNTIMECLASS_NAME
+    returnh $?
+}
+
+wait_pod_ready() {
+    local TIMEOUT_SECS=$1
+    local POD_NAME=$2
+
+    timeout $TIMEOUT_SECS bash -c '
+        pod_name=$0
+        while [ true ]
+        do
+            pod_info=$(kubectl get pods -A 2>&1 | grep $pod_name | awk '"'"'{print $3" "tolower($4)}'"'"')
+            IFS='"'"' '"'"'
+            echo $pod_info
+            read -a temp_arr <<< $pod_info
+            pod_status=${temp_arr[1]}
+            temp=${temp_arr[0]}
+            IFS="/"
+            read -a temp_arr <<< $temp
+            num_pod_ready=${temp_arr[0]}
+            num_pod_total=${temp_arr[1]}
+            
+            echo "[Debug] pod_ready_status: $num_pod_ready/$num_pod_total $pod_status"
+
+            if [[ $num_pod_ready = $num_pod_total && $pod_status = running ]]; then
+                break
+            fi
+
+            sleep 2
+        done
+    ' $POD_NAME
+
+    return $?
+}
+
+wait_runtimeclass_ready() {
+    local TIMEOUT_SECS=$1
+    local RUNTIMECLASS_NAME=$2
+
+    timeout $TIMEOUT_SECS bash -c '
+        runtimeclass_name=$0
+        while [ true ]
+        do
+            kubectl get runtimeclass 2>&1 | grep $runtimeclass_name
+            if [ $? = 0 ]; then
+                break
+            fi
+            sleep 2
+        done
+    ' $RUNTIMECLASS_NAME
+
+    return $?
+}
+
+wait_runtimeclass_deleted() {
+    local TIMEOUT_SECS=$1
+    local RUNTIMECLASS_NAME=$2
+
+    timeout $TIMEOUT_SECS bash -c '
+        runtimeclass_name=$0
+        while [ true ]
+        do
+            kubectl get runtimeclass 2>&1 | grep $runtimeclass_name
+            if [ $? != 0 ]; then
+                break
+            fi
+            sleep 2
+        done
+    ' $RUNTIMECLASS_NAME
+
+    return $?
+}
+
+wait_pod_terminating() {
+    local TIMEOUT_SECS=$1
+    local POD_NAME=$2
+
+    timeout $TIMEOUT_SECS bash -c '
+        pod_name=$0
+        while [ true ]
+        do
+            kubectl get pods -A 2>&1 | grep $pod_name
+            if [ $? != 0 ]; then
+                break
+            fi
+            sleep 2 
+        done
+    ' $POD_NAME
+
+    return $?
+}
+
+wait_pod_log() {
+    local TIMEOUT_SECS=$1
+    local POD_NAME=$2
+    local LOG_CONTENT=$3
+    
+    timeout $TIMEOUT_SECS bash -c '
+        pod_name=$0
+        log_content=$1
+        while [ true ]
+        do
+            kubectl logs $pod_name 2>&1 | grep "$log_content"
+            if [ $? == 0 ]; then
+                break
+            fi
+            sleep 2
+        done
+    ' $POD_NAME $LOG_CONTENT
+
+    return $?
+}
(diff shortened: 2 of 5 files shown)


.github/workflows/test_ci_cases.yaml
```diff
@@ -0,0 +1,42 @@
...
+          ./ci/ci_tests.sh delete_eaa_cosign_encrypted_hello_world_workload
+          sleep 10
+        
+      - name: Uninstall Enclave-CC runtimeclass payload 
+        run: ./ci/ci_tests.sh uninstall_enclave_cc_runtimeclass
+
+      - name: Uninstall CoCo operator
+        run: ./ci/ci_tests.sh uninstall_coco_operator 
+     
+
```
haosanzi, line 40: please remove these empty lines

ci/utils.sh
```diff
@@ -0,0 +1,123 @@
...
+
+wait_pod_ready() {
+    local TIMEOUT_SECS=$1
+    local POD_NAME=$2
+
+    timeout $TIMEOUT_SECS bash -c '
+        pod_name=$0
+        while [ true ]
+        do
+            pod_info=$(kubectl get pods -A 2>&1 | grep $pod_name | awk '"'"'{print $3" "tolower($4)}'"'"')
```
haosanzi, line 23: Perhaps we can check whether the pod is ready by checking whether the pod is running and in the ready state
for example:
```
          while true; do
            count=$(kubectl get pod $pod_name| grep "Running" | grep -c "1/1")
            if [ $count -eq 1 ]; then
              break
            fi
            sleep 5
          done
```

.github/workflows/test_ci_cases.yaml
```diff
@@ -0,0 +1,42 @@
...
+  CI-tests:
+    runs-on: self-hosted
+    steps: 
+
+      - name: Checkout latest enclave-cc source code
+        uses: actions/checkout@v3
+        with:
+          ref: main
+                
+      - name: Install CoCo operator
```
haosanzi, line 20: Once CI/CD fails in the intermediate step, the environment is not clean up.
When CI/CD fails, we have to log in to the machine to check the environment which can save the work of reproducing

It is recommended to clean up the environment when ci/cd starts to run.

ci/utils.sh
huliucheng1: I checked the status of the specific pod in the `awk` command, `$3` meaning n/N , and `$4` meaning containercreating/running etc. For coco operator, the ready pod number should be 2/2. And for cc runtimeclass, the ready pod number should be 1/1. For the sake of generality, I implemented it by getting the ready pod numbers automatically.

.github/workflows/test_ci_cases.yaml
huliucheng1: It is hard to clean up the environment at the current stage of this project when it failed to install pods. There are too many factors impacting whether the CI can succeed, such as the remaining space of the CI machine, the network quality and some nonexistent directory paths and etc. It can even be stuck when removing the installed pods.
Currently I just clean up the environment of CI machine manually and rerun the Github actions task if the CI failed.
Maybe the best solution is to back up the system and every time run the CI task, recover the system from backup firstly.

haosanzi (approved):

.github/workflows/test_ci_cases.yaml
```diff
@@ -0,0 +1,46 @@
...
+on: 
+  push:
+    branches: [ "main" ]
+
+  pull_request:
+    branches: [ "main" ]
+
+jobs: 
+  CI-tests:
+    runs-on: self-hosted
```
mythi, line 12:
```suggestion
    runs-on: [self-hosted, sgx]
```

.github/workflows/test_ci_cases.yaml
```diff
@@ -0,0 +1,46 @@
...
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v3
+        with:
+          ref: main
+
+      - name: Checkout PR source code
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
mythi, line 25: my experience is that one checkout for both PRs and main push is enough

ci/utils.sh
```diff
@@ -0,0 +1,123 @@
...
+
+wait_runtimeclass_ready() {
+    local TIMEOUT_SECS=$1
+    local RUNTIMECLASS_NAME=$2
+
+    timeout $TIMEOUT_SECS bash -c '
+        runtimeclass_name=$0
+        while [ true ]
+        do
+            kubectl get runtimeclass 2>&1 | grep $runtimeclass_name
```
mythi, line 55: I would prefer `kubectl wait` for all these utils helpers. See e2e.yaml: https://github.com/confidential-containers/operator/blob/6f241fbc056f0a5d9e1bd2c10b2cedc0782b99ff/.github/workflows/e2e.yaml#L51

.github/workflows/test_ci_cases.yaml
```diff
@@ -0,0 +1,46 @@
...
+                
+      - name: Install CoCo operator
+        run: ./ci/ci_tests.sh install_coco_operator
+
+      - name: Install Enclave-CC runtimeclass payload
```
mythi, line 30: one problem to think about is how to get a new payload built for testing purposes. For example, a pull request to enclave-cc repository is created that changes agent-enclave.  if we deploy "quay.io/confidential-containers/runtime-payload-ci:enclave-cc-HW-latest", it does not contain the changes we want to test.

mythi (commented):
I'd want to see the results from the self-hosted runner before merging this PR.

.github/workflows/test_ci_cases.yaml
huliucheng1: If I don't specify the lines of `with: ref: xxxx`, the checkout action will pull the source code from repo by default, rather than from PR commits.

ci/utils.sh
huliucheng1: `kubectl get xxx | grep xxxx` can print status of pods as a non-block method, so we can read changing status from Github Actions directly, and analyze which process spends abnormal time. `kubectl wait` only concerns results.

.github/workflows/test_ci_cases.yaml
huliucheng1: > I'd want to see the results from the self-hosted runner before merging this PR.

https://github.com/confidential-containers/enclave-cc/actions/runs/4080955181/jobs/7035496977

.github/workflows/test_ci_cases.yaml
huliucheng1: Let the yaml file which will push the payload image to quay.io waits for security scanning by quay.io until it succeeds. And set this ci test yaml file depends on that yaml file. I think we can open a new issue to discuss this problem in detail. : )

.github/workflows/test_ci_cases.yaml
mythi: > If I don't specify the lines of `with: ref: xxxx`, the checkout action will pull the source code from repo by default, rather than from PR commits.

only one `uses: actions/checkout@v3` should be enough.

> > I'd want to see the results from the self-hosted runner before merging this PR.
> 
> https://github.com/confidential-containers/enclave-cc/actions/runs/4080955181/jobs/7035496977

Sorry I was confused by "CI tests" which I thought refers to our original checks. Can we move this functionality under `enclave-cc e2e test`?

ci/utils.sh
mythi: > `kubectl wait` only concerns results.

if the expected condition does not trigger you can print out errors and check detailed logs

.github/workflows/test_ci_cases.yaml
mythi: we don't publish payloads for PRs but only when new code is merged in `main` so that will not work. for PRs one simple approach would be to just have a matrix job for both HW and SW

.github/workflows/test_ci_cases.yaml
huliucheng1: Do you mean move these CI tests to `https://github.com/confidential-containers/enclave-cc/blob/main/.github/workflows/e2e.yaml`, and only run CI tests after every PR is merged?

ci/utils.sh
huliucheng1: OK.

.github/workflows/test_ci_cases.yaml
huliucheng1: I don't understand. can you give a simple example scheme? When the code of the agent is changed, how can I use the latest payload?

mythi: you cannot use the payload images in a CI pipeline because they do not contain the changes you want to test. See `e2e.yaml`: we build the agent-enclave and boot-instance bundlers on the runner and install them manually.

.github/workflows/test_ci_cases.yaml
mythi: we want CI tests for every PR but given the limitations we have with the payload images through the operator based installation, I was thinking could we just reuse `e2e.yaml` based flow and create a matrix workflow for HW (runs-on self-hosted) and SW (runs-on ubuntu-xxx).

.github/workflows/test_ci_cases.yaml
huliucheng1: Do you mean I can re-use the payload image built by `e2e.yaml`, and run `e2e.yaml` on sgx machine as a HW matrix test?

huliucheng1: The sgx machine won't be re-created as a fresh new system every time CI tests begin to run. If I set `e2e.yaml` as a matrix test, many installation steps in `e2e.yaml` will be skipped, such as installing crictl, containerd, cni and etc.

mythi: > Do you mean I can re-use the payload image built by `e2e.yaml`, and run `e2e.yaml` on sgx machine as a HW matrix test?

yes that is what I had in mind but that's just one option. My key message is that for the CI testing PRs we must ensure the code changes in that PR get also tested.

For a robust CI setup, it would be good to always start a fresh setup and not expect the previous round left the system in clean state.

huliucheng1: Is it possible to use the following solution:  build a docker image in which prerequisites such as occlum runtime, containerd and k8s etc are installed. This image is built nightly and be placed at sgx ci machine. Then every time commits are updated from a PR, the ci machine will start a container from this image, and run build payloads and start a workload etc inside the container.
To use this method, we can run ci tests concurrently if there are a lot of commits at the same time. In addition, we don't need to clean up the environment because every container is fresh new.

huliucheng1: In the current stage I will add the part of building payload image locally. This is just a simple CI pipeline. And I will implement container-style CI tests at next version. How about this scheme?

mythi: that works. The "container-style CI" could follow what we use with the operator: the latest payload is tested in a nightly test: https://github.com/confidential-containers/operator/blob/6f241fbc056f0a5d9e1bd2c10b2cedc0782b99ff/.github/workflows/enclave-cc-cicd.yaml#L49-L54

huliucheng1: The `ImagePullPolicy` should be change to `IfNotPresent` in the following two lines to allow me to test payload images locally:
https://github.com/confidential-containers/operator/blob/18733db5321bf6649a56c7d698f323aecc294800/controllers/ccruntime_controller.go#L878
https://github.com/confidential-containers/operator/blob/18733db5321bf6649a56c7d698f323aecc294800/controllers/ccruntime_controller.go#L686

ariel-adam:
Moving to release V0.5.0 since V0.4.0 is closing tomorrow

mythi:
replaced by #123

PR #98, opened 2023-01-18
build: Remove $HOME/.rustup content from the final image

fidencio:
We've noticed that in the fcbf7c924d280fe328aa96de09ab9ce72bf5ccb9 commit the image has grown from ~700MB to 1.6GB.  The commit introduced some changes that, for sure, should make the image bigger, but not *that big*.

After analysing the image content, I've noticed that the whole $HOME/.rustup has been added to the image, and that's most likely the root cause of the issue.


fidencio:
Yep, just the `$HOME/.rustup` folder is 1.2GB :-)

mythi (approved):
thanks!

PR #99, opened 2023-01-18
build: Remove /enclave-cc from the final image

fidencio:
This was also added by fcbf7c924d280fe328aa96de09ab9ce72bf5ccb9 and causes a /enclave-cc directory, with the enclave-cc source code, vendors, whatnot to be left as part of the agent-instance.tar.


fidencio:
The [previous commit](https://github.com/confidential-containers/enclave-cc/pull/98) reduced the image size by 300MB, which leaves an image that's still quite big (1.3GB).

After further checking the content before the commit that caused the bump and after the commit, this seems to be the last leftover that must be removed.

mythi (approved):

PR #109, opened 2023-02-16
boot-instance: allow more images to run

mythi:
Setting rootfs_entry to /bin allowed to run binaries from that directory only. / works too and allows binaries from any path to run.

At the same time, Occlum did not get all the dynamically linked libraries loaded due to missing lib search paths. Add trusted LD_LIBRARY_PATH setting for the most common paths.

### DIFF
--- a/src/runtime-boot/init/src/main.rs
+++ b/src/runtime-boot/init/src/main.rs
@@ -22,7 +22,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let rootfs_key = b"c7-32-b3-ed-44-df-ec-7b-25-2d-9a-32-38-8d-58-61";
     let rootfs_upper_layer = "/sefs/upper";
     let rootfs_lower_layer = "/sefs/lower";
-    let rootfs_entry = "/bin";
+    let rootfs_entry = "/";
 
     // Get the key of FS image if needed
     let key = {
--- a/tools/packaging/build/boot-instance-bundle/jq.filter
+++ b/tools/packaging/build/boot-instance-bundle/jq.filter
@@ -1,3 +1,4 @@
 .resource_limits.user_space_size = "600MB" |
 .resource_limits.kernel_space_stack_size= "2MB" |
+.env.default |= . + ["LD_LIBRARY_PATH=/opt/occlum/glibc/lib:/lib/x86_64-linux-gnu/:/usr/local/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/usr/local/lib"] |
 if $ENV.SGX_MODE == "SIM" then .metadata.debuggable = true else .metadata.debuggable = false end


mythi:
@qzheng527 please help review

src/runtime-boot/init/src/main.rs
```diff
@@ -22,7 +22,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let rootfs_key = b"c7-32-b3-ed-44-df-ec-7b-25-2d-9a-32-38-8d-58-61";
     let rootfs_upper_layer = "/sefs/upper";
     let rootfs_lower_layer = "/sefs/lower";
-    let rootfs_entry = "/bin";
+    let rootfs_entry = "/";
```
qzheng527, line 25: The reason Occlum limit the entry path is for security. If changing to "/", every binaries in the image could be executed. For example, a workload " /opt/app/app_bin" saves some sensitve data in TEE which is encrypted. If no entry path limit, hacks may get the chance to run "/bin/cat" to dump the sensitve data out.
I understand that is the "trade-off" to support more images with different entry paths.
If we do that, the enclave-cc flow should take care of the security.

mythi: OK so this is the same as `entry_points`  in `Occlum.json`? 

> For example, a workload " /opt/app/app_bin" saves some sensitve data in TEE which is encrypted. If no entry path limit, hacks may get the chance to run "/bin/cat" to dump the sensitve data out.

Right, it's still open for CoCo in general so I would be OK to go with this PR. Is it possible to do this mount as read-only?

qzheng527: > OK so this is the same as `entry_points` in `Occlum.json`?
> 
> > For example, a workload " /opt/app/app_bin" saves some sensitve data in TEE which is encrypted. If no entry path limit, hacks may get the chance to run "/bin/cat" to dump the sensitve data out.
> 
> Right, it's still open for CoCo in general so I would be OK to go with this PR. Is it possible to do this mount as read-only?

Yes, it is the same as entry_points in Occlum.json.
It has nothing to do this the mount way. No limit in entry path just gives a "hole" for malicious user to do what they want. 
But I am OK with this PR in current stage.

mythi: > It has nothing to do this the mount way. No limit in entry path just gives a "hole" for malicious user to do what they want.

Yes the question about mount was not related to `entry_point` directly. I was just thinking can we make the rootfs read-only to ensure nothing sensitive is written there.

mythi: @qzheng527 could you also "Approve" via github review, thanks

qzheng527 (approved):

haosanzi (approved):

src/runtime-boot/init/src/main.rs
dcmiddle: > If we do that, the enclave-cc flow should take care of the security.

@qzheng527 could you say a little more about what would be expected of the enclave-cc flow to enforce security?

mythi:
I would like this to be in `v0.4.0`

src/runtime-boot/init/src/main.rs
mythi: > > If we do that, the enclave-cc flow should take care of the security.
> 
> @qzheng527 could you say a little more about what would be expected of the enclave-cc flow to enforce security?

@qzheng527 did you see this question?

qzheng527: > > If we do that, the enclave-cc flow should take care of the security.
> 
> @qzheng527 could you say a little more about what would be expected of the enclave-cc flow to enforce security?

@dcmiddle @mythi It is a open question, I don't have solid solution for this. But perhaps, the enclave-cc flow can handle the parse of the container image, then pass the entrypoint plus other information (key, envs, ...) to the app enclave to boot.

dcmiddle: @qzheng527 thanks for the clarification. I opened an issue so we can collect thoughts.
https://github.com/confidential-containers/enclave-cc/issues/119

PR #110, opened 2023-02-22
enclave-agent: Bump image-rs to 948db858579be538724266a650e840e75e7e824d

fidencio:
That's the image-rs version that will become the v0.4.0, and the main change that affects Enclave CC is the way we build it, as we can simply use the `enclave-cc` feature.


Xynnn007:
@fidencio can you try ae0a105da17958b09e866b631faf6fbac5e3fd18 again?

haosanzi (approved):

PR #118, opened 2023-03-01
Bump image-rs to v0.4.0 tag

fidencio:


dcmiddle (approved):
Per clarification from @fidencio this tag is at the commit we are already using.
i.e.
```
(main)$ git show-ref v0.4.0
948db858579be538724266a650e840e75e7e824d refs/tags/v0.4.0
```

PR #127, opened 2023-03-09
enclave-agent: enable signature-cosign image-rs feature

mythi:
Commit c448ebc9 moved to a new image-rs that had re-organized its features. The new "enclave-cc" feature does not enable signature-cosign by default but our default test images depend on it so enable it.

/cc @Xynnn007


Xynnn007 (approved):
LGTM. When the image-rs side is updated, I'll come back to sync the `Cargo.toml`

dcmiddle (approved):

PR #132, opened 2023-03-13
payload: Use ubuntu 20.04 image as base

fidencio:
Instead of using CentOS as the base image, let's follow what's been done for the preInstall / postInstall / Kata Containers payload, and switch to using an Ubuntu image instead.

### DIFF
--- a/tools/packaging/build/Dockerfile
+++ b/tools/packaging/build/Dockerfile
@@ -1,24 +1,24 @@
 ARG IMAGE
-FROM ${IMAGE:-registry.centos.org/centos}:7
+FROM ${IMAGE:-ubuntu}:20.04
 ARG ENCLAVE_CC_ARTIFACTS=./enclave-cc-static.tar.xz
 ARG DESTINATION=/opt/enclave-cc-artifacts
 
 COPY ${ENCLAVE_CC_ARTIFACTS} ${WORKDIR}
 
+ENV DEBIAN_FRONTEND=noninteractive
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
 RUN \
-yum -y update && \
-yum -y install xz && \
-yum clean all && \
+apt-get update && \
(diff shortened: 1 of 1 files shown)


tools/packaging/build/Dockerfile
```diff
@@ -1,24 +1,24 @@
...
-yum -y install xz && \
-yum clean all && \
+apt-get update && \
+apt-get install -y --no-install-recommends apt-transport-https ca-certificates curl xz-utils systemd && \
+mkdir -p /etc/apt/keyrings/ && \
+curl -fsSLo /etc/apt/keyrings/kubernetes-archive-keyring.gpg https://packages.cloud.google.com/apt/doc/apt-key.gpg && \
+echo "deb [signed-by=/etc/apt/keyrings/kubernetes-archive-keyring.gpg] https://apt.kubernetes.io/ kubernetes-xenial main" | tee /etc/apt/sources.list.d/kubernetes.list && \
+apt-get update && \
+apt-get install -y --no-install-recommends kubectl && \
+apt-get clean && rm -rf /var/lib/apt/lists/ && \
```
mythi, line 19: since we only care about `kubectl`, would it be faster if we just install the binary:

```
$  curl -sLfS https://dl.k8s.io/v1.26.2/kubernetes-client-linux-amd64.tar.gz | tar tz
kubernetes/
kubernetes/client/
kubernetes/client/bin/
kubernetes/client/bin/kubectl
kubernetes/client/bin/kubectl-convert
```

fidencio: I totally agree, @mythi.  Do you mind if I address this as a follow-up PR?

I'll need to change this in a few different places:
* preInstall payload
* Kata Containers
* here

mythi: a follow-up PR is fine.

mythi (approved):

PR #139, opened 2023-03-29
dependabot: ignore containerd/ttrpc v1.2.x

mythi:
It requires more work to get this working so keep using v1.1.x.


fidencio (approved):
lgtm, thanks @mythi!

Issue #130, opened 2023-03-11
Roadmap to support new image format for eaa-kbc


Xynnn007:
As stated https://github.com/confidential-containers/enclave-cc/issues/108, we brought a unified format for encrypted images and a new mechanism to identify kbs resources like decryption keys. We need the following jobs to be done to support the enhancement for encrypted image.

##### For eaa-kbc

- [x] Implement verdictd to support the new resource uri scheme
https://github.com/inclavare-containers/verdictd/issues/29
https://github.com/confidential-containers/attestation-agent/pull/152
- [x] update the image-rs version for enclave-agent
- [x] prepare new test images
https://github.com/confidential-containers/enclave-cc/issues/108
- [x] update ci for eaa-kbc

mythi:
@Xynnn007 thanks for submitting this! For tracking purposes, I was wondering what we need for enclave-cc. The new image format is implicitly supported via #120 and for that we also need #108, right? With that, maybe this issue should just track "support new image format with eaa_kbc in enclave-cc"?

Xynnn007:
> @Xynnn007 thanks for submitting this! For tracking purposes, I was wondering what we need for enclave-cc. The new image format is implicitly supported via #120 and for that we also need #108, right? With that, maybe this issue should just track "support new image format with eaa_kbc in enclave-cc"?

Oh, yes, they are duplicated. Let me fix this

PR #140, opened 2023-03-29
Bring new test images for CI

Xynnn007:
Fixes: #130

### DIFF
--- a/src/enclave-agent/Cargo.toml
+++ b/src/enclave-agent/Cargo.toml
@@ -13,7 +13,7 @@ clap = "2.33.3"
 # logger module
 env_logger = "0.10.0"
 
-image-rs = { git = "https://github.com/confidential-containers/image-rs", default-features = false, rev = "v0.4.0" }
+image-rs = { git = "https://github.com/confidential-containers/image-rs", default-features = false, rev = "cc626a9" }
 kata-sys-util = { git = "https://github.com/kata-containers/kata-containers", rev = "4b57c04c3379d6adc7f440d156f0e4c42ac157df" }
 log = "0.4.11"
 protocols = { path = "../libs/protocols" }
@@ -25,8 +25,11 @@ tokio = { version = "1.14.0", features = ["rt-multi-thread", "macros", "signal"]
 
 ttrpc = { git = "https://github.com/haosanzi/ttrpc-rust", features = ["async"] }
 
+[patch.crates-io]
+oci-distribution = { git = "https://github.com/krustlet/oci-distribution.git", rev = "f44124c" }
+
 [features]
-default = ["image-rs/enclave-cc", "image-rs/signature-cosign"]
+default = ["image-rs/enclave-cc"]
 simple = ["image-rs/snapshot-unionfs", "image-rs/signature-cosign", "image-rs/signature-simple", "image-rs/keywrap-native", "image-rs/encryption"]
 
 [build-dependencies]
--- a/test/e2e/hello-world-encrypted-HW.json
+++ b/test/e2e/hello-world-encrypted-HW.json
@@ -3,7 +3,7 @@
     "name": "hello-world-encrypted-container"
   },
   "image":{
-    "image": "docker.io/eqmcc/helloworld_enc"
+    "image": "ghcr.io/confidential-containers/test-container-enclave-cc:encrypted"
   },
   "envs": [
     {"key": "OCCLUM_RELEASE_ENCLAVE", "value": "1"}
--- a/test/e2e/hello-world-encrypted-SIM.json
+++ b/test/e2e/hello-world-encrypted-SIM.json
@@ -3,7 +3,7 @@
     "name": "hello-world-encrypted-container"
   },
   "image":{
-    "image": "docker.io/mylinen/hello-world-encrypted:samplekbc2"
+    "image": "ghcr.io/confidential-containers/test-container-enclave-cc:sample_kbc_encrypted"
   },
   "envs": [
     {"key": "OCCLUM_RELEASE_ENCLAVE", "value": "0"}
--- a/src/enclave-agent/Cargo.lock
+++ b/src/enclave-agent/Cargo.lock
@@ -206,17 +206,22 @@ checksum = "2ce4f10ea3abcd6617873bae9f91d1c5332b4a778bd9ce34d0cd517474c1de82"
 [[package]]
 name = "attestation_agent"
 version = "0.1.0"
-source = "git+https://github.com/confidential-containers/attestation-agent?rev=b45b0f8#b45b0f81b9671588b4634feb515bc484f3346922"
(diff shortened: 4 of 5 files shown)


src/enclave-agent/Cargo.toml
```diff
@@ -13,7 +13,7 @@ clap = "2.33.3"
 # logger module
 env_logger = "0.10.0"
 
-image-rs = { git = "https://github.com/confidential-containers/image-rs", default-features = false, rev = "v0.4.0" }
+image-rs = { git = "https://github.com/confidential-containers/image-rs", features = ["enclave-cc"], default-features = false, rev = "cc626a9" }
```
mythi, line 16: The `features` addition here is causing the RATS-TLS build issues:

```suggestion
image-rs = { git = "https://github.com/confidential-containers/image-rs", default-features = false, rev = "cc626a9" }
```

Xynnn007: Thanks, I've updated

test/e2e/hello-world-encrypted-SIM.json
```diff
@@ -3,7 +3,7 @@
     "name": "hello-world-encrypted-container"
   },
   "image":{
-    "image": "docker.io/mylinen/hello-world-encrypted:samplekbc2"
+    "image": "docker.io/xynnn007/helloworld:sample_kbc_enc"
```
mythi, line 6: As part of https://github.com/kata-containers/tests/issues/5521, my goal is to be able to get rid of personal images and have all of the test images in the same place.  Now that the tests for this PR are green, can we get the images to moved to the same place where the rest of the test coco images?

Xynnn007: Agreed. Now enclave-cc uses two test encrypted images
- decrypted by sample-kbc in SIM: actually `sample-kbc` is not used anymore for kata-cc
- decrypted by eaa-kbc (replace with cc-kbc in future): I suggest to replace with [ccv0-ssh](https://hub.docker.com/layers/xynnn007/ccv0-ssh/encrypted/images/sha256-ac0ed3364d54120a1025c74f555b21cb378d4d0f62a398c5ce3e1b89fa4ca637?context=repo) (which I have not test yet, though), which will be moved to ghcr.io in HW

Xynnn007:
I have replaced the test image with the one of community. However, that one is not signed by cosign now. I will use the policy 
```
{
    "default": [{"type": "reject"}], 
    "transports": {
        "docker": {
            "ghcr.io/confidential-containers": [
                {
                    "type": "insecureAcceptAnything"
                }
            ]
        }
    }
}
```

which means the image will be accepted without signature verification.

test/e2e/hello-world-encrypted-HW.json
```diff
@@ -1,19 +1,19 @@
...
-    "image": "docker.io/eqmcc/helloworld_enc"
+    "image": "ghcr.io/confidential-containers/test-container:encrypted"
   },
   "envs": [
     {"key": "OCCLUM_RELEASE_ENCLAVE", "value": "1"}
   ],
   "command": [
     "/run/rune/boot_instance/build/bin/occlum-run",
-    "/bin/hello_world"
+    "/usr/sbin/sshd"
```
mythi, line 13: is the image running OK? The logs report:
```
sshd re-exec requires execution with an absolute path
["init"]
```

mythi (commented):
only some minor comments. Can you squash some of the commits (e.g., `image-rs` + Cargo updates)

test/e2e/hello-world-encrypted-SIM.json
mythi: @Xynnn007 Do you think it makes sense to continue using `sample-kbc` based image for SIM? (btw, we have this `hello` image used in operator e2e too: https://github.com/confidential-containers/operator/blob/88a0a60b47bf76dcc9a37452a53063df0917a37c/tests/e2e/enclave-cc-pod-sim.yaml#L9)

Xynnn007: I suggest not. We can use `offline-fs-kbc` instead to test key brokering without remote attestation -- which means that we should get rid of `sample-kbc`. How do you feel?

src/enclave-agent/Cargo.toml
```diff
@@ -25,8 +25,11 @@ tokio = { version = "1.14.0", features = ["rt-multi-thread", "macros", "signal"]
 
 ttrpc = { git = "https://github.com/haosanzi/ttrpc-rust", features = ["async"] }
 
+[patch.crates-io]
```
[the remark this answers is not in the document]
Xynnn007: We added this patch inside image-rs, but it does not work for downstream projects, like here and kata-agent. We need to re-patch here.

test/e2e/hello-world-encrypted-HW.json
Xynnn007: No. It's not ok. `sshd` requires absolute path to run like `/usr/sbin/sshd` rather than `./sshd`. This might be caused by `occlum-run`?  Could you help to give any comments? @qzheng527

test/e2e/hello-world-encrypted-SIM.json
mythi: Does it mean the decryption keys must be copied in the occlum instance?

Xynnn007: no, we do not. we just need to set up a file `/etc/offline-fs-kbc-keys.json` on the ci machine like [this](https://github.com/confidential-containers/image-rs/blob/main/test_data/offline-fs-kbc/aa-offline_fs_kbc-keys.json). Then enclave-agent with `attestation_agent/offline_fs_kbc` enabled can look up for keys from that file.

But seems that the code freeze is near, we can just use this sample kbc encrypted image, and start to change next release?

I think it is not a big problem

test/e2e/hello-world-encrypted-HW.json
Xynnn007: This is because occlum will change argv[0] from a absolute path into the binary name here https://github.com/occlum/occlum/blob/master/src/run/main.c#L24 . The error is not related to enclave-cc, but it seems that we can not realize a successfully running `ccv0-sshd` service via enclave-cc for v0.5.0.

Two choices for us:
1. Use the previous `hello_world` instead
2. Use this image, and treat it is launched "successfully"

Xynnn007: > I'd prefer option 1.

I have pushed the encrypted image (hello-world) to `ghcr.io`. However, I do not have permission to make it public. Let me ask Tobin for help and we can retrigger HW test then

mythi: @Xynnn007 OK, thanks! I think the PR looks good otherwise.

mythi:
@Xynnn007 thanks!

PR #144, opened 2023-04-14
deps: Bump image-rs to v0.5.1

fidencio:
Let's update image-rs to the latest released tag.


wainersm (approved):
The change is correct but I do not understand why CI failed

dcmiddle (approved):

Issue #120, opened 2023-03-02
Roadmap for enclave-cc to support CoCo Key Broker System


Xynnn007:
### Background

Now enclave-cc uses eaa-kbc & verdictd as underlying attestation & confidential resource broker componant. At the same time, CoCo key broker system is under development. CoCo Key Broker System includes 
- [CC-KBC in Attestation-Agent](https://github.com/confidential-containers/attestation-agent/tree/main/src/kbc_modules/cc_kbc): To collect hardware & software evidence and do remote attestation and make resource request from AS
- [Attestation-Service](https://github.com/confidential-containers/attestation-service): To verify the evidence sent by CC-KBC
- [CC-KBS](https://github.com/confidential-containers/kbs): To distribute confidential resources

Currently, we also use eaa-kbc & verdictd, where
- [eaa-kbc](https://github.com/confidential-containers/attestation-agent/tree/main/src/kbc_modules/eaa_kbc): same functionality as cc-kbc
- [verdictd](https://github.com/inclavare-containers/verdictd): same functionality as AS + CC-KBS

### What we want

To support CoCo key broker system in enclave-cc, we need the following things. Let's make a simple roadmap for this
- [x] Implement Occlum attester in Attestation-agent's cc-kbc
https://github.com/confidential-containers/attestation-agent/issues/134
- [x] Implement SGX verifier in Attestation-Service (As Occlum & Gramine share sgx qvl)
https://github.com/confidential-containers/attestation-service/issues/51
- [x] Add build features for both `attestation-agent/eaa-kbc` and `attestation-agent/cc-kbc` for agent-enclave
- [x] Related tests/ci using coco-key broker system

PR #146, opened 2023-05-04
Feat cckbc

Xynnn007:
Related to https://github.com/confidential-containers/enclave-cc/issues/120

This PR:
- Added Add build features for both `attestation-agent/eaa-kbc` and `attestation-agent/cc-kbc` for agent-enclave
- Update CI to cover cc-kbc & occlum


tools/packaging/build/agent-enclave-bundle/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM ubuntu:20.04 as builder
 
 ARG RATS_TLS_REV=5de6fc308f2a18f4105429451c85e6173b42ab9d
+ARG KBC=cc-kbc
```
mythi, line 4: Can you verify that the payload images we publish continue to work, e.g., in the operator CI where we test with the SIM environment and simplekbc.

Also, if we default to this, can we stop building RATS-TLS to the occlum instance

mythi (commented):
This is nice work. I only have one comment about what we publish as default for HW and SIM

tools/packaging/build/agent-enclave-bundle/Dockerfile
Xynnn007: > Can you verify that the payload images we publish continue to work, e.g., in the operator CI where we test with the SIM environment and simplekbc.

Sorry, I am not familiar with this. Could you help with this? I've updated the payload publishment CI in the following commits.

> Also, if we default to this, can we stop building RATS-TLS to the occlum instance
Good point! I've update the Dockerfile to have a conditional build depending on env `KBC`

src/enclave-agent/Cargo.toml
```diff
@@ -29,8 +29,18 @@ ttrpc = { git = "https://github.com/haosanzi/ttrpc-rust", features = ["async"] }
 oci-distribution = { git = "https://github.com/krustlet/oci-distribution.git", rev = "f44124c" }
 
 [features]
-default = ["image-rs/enclave-cc"]
-simple = ["image-rs/snapshot-unionfs", "image-rs/signature-cosign", "image-rs/signature-simple", "image-rs/keywrap-native", "image-rs/encryption"]
+default = ["cc-kbc-rustls-tls"]
+
+# confidential resource injection using sample-kbc
+simple = ["image-rs/snapshot-unionfs", "image-rs/signature-cosign-rustls", "image-rs/signature-simple", "image-rs/keywrap-native", "image-rs/encryption-ring", "image-rs/oci-distribution-rustls"]
+
```
mythi, line 36: I created this dummy `simple` target just to be able to run CI tests without having RATS-TLS installed. We can probably drop this completely and move the Rust CI checks to use `cc-kbc-*`feature instead.

`sample-kbc` is always enabled, btw.

src/enclave-agent/Makefile
```diff
@@ -7,8 +7,14 @@ else
...
-all:
-	RUSTFLAGS="-C link-args=-Wl,-rpath=/usr/local/lib/rats-tls:/usr/local/lib" cargo build $(release)
+all cc-kbc:
+	RUSTFLAGS="-C link-args=-W" cargo build $(release)
+
+eaa-kbc:
+	RUSTFLAGS="-C link-args=-Wl,-rpath=/usr/local/lib/rats-tls:/usr/local/lib" cargo build $(release) --no-default-features --features eaa-kbc-rustls-tls
+
+sample-kbc:
+	RUSTFLAGS="-C link-args=-W" cargo build $(release) --no-default-features --features simple
```
mythi, line 17: perhaps we can drop this because it's not needed

src/enclave-agent/Cargo.toml
Xynnn007: > sample-kbc is always enabled, btw.

Yes. Now a separate `simple` feature and a `sample-kbc` target helps to build an enclave-agent binary that does not need to run in true sgx libos, though I am ok either abondon `simple` or not

Xynnn007: FYI: if we abondon `sample-kbc`, we might not need a SIM test in CI. Instead we only need two, s.t. `cc-kbc` and `eaa-kbc`

mythi: Good point about `simple` not requiring the LibOS, let's just keep it then. I prefer to have `sample_kbc` available because it allows us to run basic CI checks using Github actions without additional setup steps, like in the operator repo: https://github.com/confidential-containers/operator/blob/fe952c8b65e9ab76fc6e963e83caf1a7310627c5/.github/workflows/e2e.yaml#L54-L63

mythi (approved):

PR #150, opened 2023-05-17
agent-enclave: bump image-rs to v0.5.2

Xynnn007:


src/enclave-agent/Cargo.toml
```diff
@@ -13,7 +13,7 @@ clap = "2.33.3"
 # logger module
 env_logger = "0.10.0"
 
-image-rs = { git = "https://github.com/confidential-containers/image-rs.git", default-features = false, rev = "44f27c2" }
+image-rs = { git = "https://github.com/Xynnn007/image-rs.git", default-features = false, rev = "c191a12" }
```
mythi, line 16: perhaps we can move to `image-rs` latest?

Xynnn007: I think we do not need this PR anymore. We can wait until the release

PR #165, opened 2023-05-31
ci: abondon eaa-kbc for ci test

Xynnn007:
Let's remove eaa-kbc from CI as cc-kbc is supported now.


Xynnn007:
Do we need to delete the corresponding files, s.t. `hello-world-encrypted-HW-eaa-kbc.json` etc in this PR?

mythi:
> Do we need to delete the corresponding files, s.t. `hello-world-encrypted-HW-eaa-kbc.json` etc in this PR? I think some old school players might want to play with eaa-kbc, for which leaving those files still in the code seems good.

We can keep that for a while.

.github/workflows/e2e.yaml
```diff
@@ -19,9 +19,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - runner: sgx-ecc
-            sgx_mode: HW
-            kbc: eaa-kbc
```
mythi, line 24: please check `payload.yaml` (and `release.yaml`) too.

Xynnn007: Fixed in `payload.yaml`. `release.yaml` looks good, but I'm not sure whether the  name changing `quay.io/confidential-containers/runtime-payload:enclave-cc-HW-cc-kbc` still works?

mythi (approved):

PR #166, opened 2023-06-01
packaging: fix payload build

Xynnn007:
As
https://github.com/kubernetes/release/issues/2862#issuecomment-1565660554, public key release of kubernetes have been changed.


mythi (approved):
/cc @fidencio

PR #170, opened 2023-06-07
Workflows: fix release workflow

fitzthum:
Should the release workflow match the push workflow? Currently we are releasing a `SIM-cc-kbc` bundle, which does not match our CI bundles. Is this necessary? 

@Xynnn007 @mythi @dcmiddle

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,9 +9,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        asset:
-          - HW
-          - SIM
+        include:
+          - sgx_mode: SIM
+            kbc: sample-kbc
+          - sgx_mode: HW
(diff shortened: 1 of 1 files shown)


dcmiddle (approved):
This was the conclusion I was reaching too. 
When the release script doesn't set the KBC variable it gets defaulted to cc-kbc in a few places including the label string for the build.

Xynnn007 (approved):
Oh yes. This was a mistake that KBC is not defined. Thanks for fixing this!

mythi:
I've been offline this week so I can only comment after the fact.

This change is mostly cosmetic and not necessary for the release since we dropped eaa-kbc. KBC in the bundle name is a leftover and likely going away from the push workflow too.

We rely on sample_kbc being enabled by default in AA. With this change the problem with the release is that SIM bundle does not have cc-kbc enabled. HW has both.

PR #175, opened 2023-06-15
build: lock Occlum version

mythi:
At the same time, it was noticed occlum-sgx-tools is better to pull _sim.so libraries from instead of separately installing the SDK package.


fidencio (approved):
lgtm, thanks @mythi!

PR #176, opened 2023-06-15
workflows: fix payload push trigger syntax

mythi:
dependabot pushes to the repository trigger payload.yaml workflows which should not happen at all.

It's likely that Github does not like the branch names without hyphens and the filter gets broken.

### DIFF
--- a/.github/workflows/payload.yaml
+++ b/.github/workflows/payload.yaml
@@ -1,8 +1,8 @@
 name: Publish Enclave CC CI payloads for Confidential Containers
 on:
   push:
-    branch:
-      - main
+    branches:
+      - 'main'
 
 jobs:
   build-asset:


fidencio (approved):
lgtm, thanks @mythi!

fidencio (changes requested):
Actually, let me take a step back here. :-)

.github/workflows/payload.yaml
```diff
@@ -2,7 +2,7 @@ name: Publish Enclave CC CI payloads for Confidential Containers
 on:
   push:
     branch:
-      - main
+      - 'main'
```
fidencio, line 5: I think both 'main' and main should work, but I think `branch` should actually be `branches`, as shown here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-filters

mythi: thanks, I missed that! fixed.

fidencio (approved):
lgtm, thanks @mythi!

PR #195, opened 2023-08-06
ccruntime: Manage runtime class creation / deletion

fidencio:
Kata Containers moved to managing the runtime class creation / deletion, leading to making the Operator "dumber", removing all the runtime class management from there.

Let's adapt to it on the Enclave CC side as well.

### DIFF
--- a/tools/packaging/deploy/enclave-cc-deploy.sh
+++ b/tools/packaging/deploy/enclave-cc-deploy.sh
@@ -25,6 +25,16 @@ function print_usage() {
 	echo "Usage: $0 [install/cleanup/reset]"
 }
 
+function create_runtimeclass() {
+	echo "Creating the runtime classes"
+	kubectl apply -f /runtimeclass/enclave-cc.yaml
+}
+
+function delete_runtimeclass() {
+	echo "Deleting the runtime classes"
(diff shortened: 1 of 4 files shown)


dcmiddle (approved):
LGTM Thanks!

tools/packaging/deploy/enclave-cc-deploy.sh
```diff
@@ -25,6 +25,16 @@ function print_usage() {
 	echo "Usage: $0 [install/cleanup/reset]"
 }
 
+function create_runtimeclass() {
+	echo "Creating the runtime classes"
+	kubectl apply -f runtimeclass/enclave-cc.yaml
```
mythi, line 30: I built the payload image and I *think* the path is not correct since `enclave-cc-deploy.sh` is under `/scripts/` and `enclave-cc.yaml` is `/runtimeclass`.

mythi, line 30: this would work:
```suggestion
	kubectl apply -f /runtimeclass/enclave-cc.yaml
```

fidencio: Changed here and also in the delete part.

PR #199, opened 2023-08-09
deploy: Use host systemctl

jepio:
Adapt enclave-cc to rely on host's systemctl by using nsenter. This change has already been done in kata-containers and is coming to CoCo operator too.

Fixes: #7520


tools/packaging/deploy/ccruntime.yaml
```diff
@@ -23,10 +23,6 @@ spec:
         name: enclave-cc-conf
       - mountPath: /opt/confidential-containers/
         name: enclave-cc-artifacts
-      - mountPath: /var/run/dbus
```
mythi, line 26: `ccruntime.yaml` in this repo is unused so the changes can be dropped (unless you don't mind removing it in a separate commit here).

mythi (commented):
thanks, LGTM but one minor comment about `ccruntime.yaml`

tools/packaging/deploy/ccruntime.yaml
fidencio: @mythi, I've updated @jepio's PR to drop the file.

PR #201, opened 2023-08-09
enclave-cc-deploy: Fix runtimeclass path

fidencio:
Instead of using `/runtimeclass/enclave-cc.yaml`, let's use the full path for it `/opt/enclave-cc-artifacts/runtimeclass/enclave-cc.yaml`

This will fix the following error faced on the Operator CI side:
```
ubuntu@operator:~/operator$ kubectl -n confidential-containers-system logs cc-operator-daemon-install-j4fxn -c cc-runtime-install-pod
copying enclave-cc artifacts onto host
Add enclave-cc as a supported runtime for containerd
Configuration exists for plugins."io.containerd.grpc.v1.cri".containerd.runtimes.enclave-cc, overwriting
Creating the runtime classes
error: the path "/runtimeclass/enclave-cc.yaml" does not exist
```

### DIFF
--- a/tools/packaging/deploy/enclave-cc-deploy.sh
+++ b/tools/packaging/deploy/enclave-cc-deploy.sh
@@ -27,12 +27,12 @@ function print_usage() {
 
 function create_runtimeclass() {
 	echo "Creating the runtime classes"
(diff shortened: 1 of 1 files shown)


mythi (approved):
thanks!

Issue #158, opened 2023-05-25
enable dependabot updates for rust dependencies


mythi:
- [x] enclave-agent
- [x] runtime-boot

mythi:
Fixed by #314

PR #208, opened 2023-09-07
dependabot: check enclave-agent dependencies

mythi:
Related: #158

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,14 @@
 version: 2
 updates:
+  - package-ecosystem: "cargo"
+    directory: "/src/enclave-agent"
+    schedule:
+      interval: daily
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,5 +1,14 @@
 version: 2
 updates:
+  - package-ecosystem: "cargo"
+    directory: "/src/enclave-agent"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 1
```
dcmiddle, line 7: Is this just so we don't get overwhelmed when turning it on for the first time?

mythi: Mainly because it is like this in the kbs repo where I copied this from but I think it makes sense to have things updated one by one to avoid extra  dependabot rebasing

PR #249, opened 2023-10-13
enclave-agent: move to clap v3 Derive API

mythi:
clap v2.33.3 is three years old and automated dependabot updates won't work anymore.

Rework enclave-agent command line args handling to the latest Derive API from clap and move to v4.4.6.

### DIFF
--- a/src/enclave-agent/src/config.rs
+++ b/src/enclave-agent/src/config.rs
@@ -1,8 +1,6 @@
-use anyhow::anyhow;
+use anyhow::{anyhow, bail};
 use serde::{Deserialize, Serialize};
 use std::{fs::File, path::Path};
-
-pub const DEFAULT_OCICRYPT_CONFIG_PATH: &str = "/etc/ocicrypt.conf";
 
 #[derive(Serialize, Deserialize, Default)]
 pub struct DecryptConfig {
@@ -33,7 +31,7 @@ pub struct OcicryptConfig {
 impl OcicryptConfig {
     pub fn new(path: String) -> Result<Self, anyhow::Error> {
         if !Path::new(&path).exists() {
-            return Err(anyhow!("ocicrypt config not found"));
+            bail!("ocicrypt config {path} does not exist.");
         }
         Ok(Self { config_path: path })
     }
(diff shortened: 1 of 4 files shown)


src/enclave-agent/src/main.rs
```diff
@@ -1,90 +1,60 @@
...
-    };
+    let cli = Cli::parse();
 
-    let dc = if let Some(file_path) = matches.value_of("decrypt-config") {
-        DecryptConfig::load_from_file(&file_path.to_string())?
-    } else {
-        DecryptConfig::default()
+    let dc = match DecryptConfig::load_from_file(&cli.decrypt_config) {
+        Ok(config) => config,
+        _ => DecryptConfig::default(),
```
Xynnn007, line 40: Do we need to raise a warning or info here to tell the caller that the given `DecryptConfig` cannot be loaded and downgrade to a default one?

mythi: makes sense, will change

mythi: updated.

src/enclave-agent/src/config.rs
```diff
@@ -33,7 +31,7 @@ pub struct OcicryptConfig {
 impl OcicryptConfig {
     pub fn new(path: String) -> Result<Self, anyhow::Error> {
         if !Path::new(&path).exists() {
-            return Err(anyhow!("ocicrypt config not found"));
+            return Err(anyhow!("ocicrypt config {} does not exist.", path));
```
Xynnn007, line 34:
```suggestion
            bail!("ocicrypt config {path} does not exist.");
```

src/enclave-agent/src/main.rs
```diff
@@ -1,90 +1,63 @@
...
+    let cli = Cli::parse();
 
-    let dc = if let Some(file_path) = matches.value_of("decrypt-config") {
-        DecryptConfig::load_from_file(&file_path.to_string())?
-    } else {
-        DecryptConfig::default()
+    let dc = match DecryptConfig::load_from_file(&cli.decrypt_config) {
+        Ok(config) => config,
+        Err(e) => {
+            println!("Setting default DecryptConfig because: {:?}", e);
```
Xynnn007, line 41: Shall we use `warn!("...")` here? As we already import dep `log` now.

mythi:
@Xynnn007 looks OK now?

Xynnn007 (approved):

PR #271, opened 2023-11-01
build: move to Go 1.21.x

mythi:


src/shim/runtime/v2/rune/v2/agent.go
```diff
@@ -94,7 +94,7 @@ func (c *agent) connect(ctx context.Context) error {
 	return nil
 }
 
-func (c *agent) disconnect(ctx context.Context) error {
+func (c *agent) disconnect(context.Context) error {
```
dcmiddle, line 97: I don't know what's stylistically recommended but I prefer the _  as in the installReqFunc declaration below. 
I guess either way is correct but we should probably stick with one.

dcmiddle (approved):
lgtm w/ the exception of one stylistic nit

dcmiddle (approved):

Issue #149, opened 2023-05-11
create rootfs_key dynamically and seal it


mythi:
We have been waiting for #20 but in the mean time, let's work on something simpler to get rid of the static `rootfs_key`.

The proposal is to create `rootfs_key` dynamically and seal with with MRSIGNER key from Occlum's getkey `ioctl()`.

Steps:

- [x] #126 
- [x] #114
- [ ] update image-rs Occlum snapshotter to make the key available under the `containerd_id` path
- [ ] update runtime boot to get the key from `containerd_id` path

Xynnn007:
Let me try to understand this:

1. Enclave-agent generates a key randomly, seal the key and then put the encrypted key to `/run/enclave-cc/<containerd_id>`
2. runtime boot get the key from `/run/enclave-cc/<containerd_id>` and unseal it

This implies that the enclave-agent and runtime boot are signed by the same signer, right?

mythi:
@Xynnn007 yes your summary is correct

ariel-adam:
@mythi will this converge for release 0.7.0 (feature freeze 12th of July) or should we move this to 0.8.0?

PR #266, opened 2023-10-30
Random Key

piotrpalcz:
This PR is related to image-rs PR https://github.com/confidential-containers/guest-components/pull/385 and Shim/rune PR https://github.com/confidential-containers/enclave-cc/pull/256

Makes Runtime boot use key stored in file by image-rs/enclave agent.

After merging and fixes next step will be to seal/unseal the file with SGX

PR #312, opened 2023-12-19
Runtime boot using a randomly generated key from image-rs passed inside enclave sefs

piotrpalcz:
This PR is related to image-rs PR https://github.com/confidential-containers/guest-components/pull/385 and replaces 
https://github.com/confidential-containers/enclave-cc/pull/266 for easier reading
After merge of image-rs PR, branch in Cargo.toml will be reverted

Makes Runtime boot use key stored in file by image-rs/enclave agent.

After merging and fixes next step will be to seal/unseal the file with SGX

Fixes: #149

### DIFF
--- a/src/runtime-boot/init/src/main.rs
+++ b/src/runtime-boot/init/src/main.rs
@@ -4,74 +4,90 @@ extern crate serde_json;
 
 use libc::syscall;
 
+use nix::mount::MsFlags;
 use std::env;
 use std::error::Error;
+use std::fs;
 use std::fs::File;
 use std::io::prelude::*;
 use std::io::{ErrorKind, Read};
 
+use anyhow::Result;
 use std::ffi::CString;
 use std::mem::size_of;
+use std::path::Path;
 
 fn main() -> Result<(), Box<dyn Error>> {
-    // TODO: Get the rootfs key and other parameters through RA/LA or PAL
-    let rootfs_key = b"c7-32-b3-ed-44-df-ec-7b-25-2d-9a-32-38-8d-58-61";
-    let rootfs_upper_layer = "/sefs/upper";
-    let rootfs_lower_layer = "/sefs/lower";
-    let rootfs_entry = "/";
-
-    // Get the key of FS image if needed
-    let key = {
-        const IMAGE_KEY_FILE: &str = "/etc/image_key";
-        // TODO: Get the key through RA or LA
-        let mut file = File::create(IMAGE_KEY_FILE)?;
-        // Writes key.
-        file.write_all(rootfs_key)?;
-
-        let key_str = load_key(IMAGE_KEY_FILE)?;
-        let mut key: sgx_key_128bit_t = Default::default();
-        parse_str_to_bytes(&key_str, &mut key)?;
-        Some(key)
-    };
-
-    let key_ptr = key
-        .as_ref()
-        .map(|key| key as *const sgx_key_128bit_t)
-        .unwrap_or(std::ptr::null());
+    let agent_boot = matches!(env::var("ENCLAVE_AGENT"), Ok(val) if val == "true" || val == "TRUE" || val == "1");
 
     // Mount the image
     const SYS_MOUNT_FS: i64 = 363;
 
-    // Example envs. must end with null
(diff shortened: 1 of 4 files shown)


src/runtime-boot/init/src/main.rs
```diff
@@ -4,74 +4,96 @@ extern crate serde_json;
...
+        false => {
+            let rootfs_upper_layer = "/sefs/upper";
+            let rootfs_lower_layer = "/sefs/lower";
+            let rootfs_entry = "/";
+
+            let fs_type = String::from("sefs");
+            let source = Path::new(&fs_type);
+
+            // Create and mount directory for fs key from agent enclave
+            fs::create_dir("/mnt")?;
```
mythi, line 40: drop?

src/runtime-boot/init/src/main.rs
```diff
@@ -4,74 +4,96 @@ extern crate serde_json;
...
+            let options = format!(
+                "dir={}",
+                Path::new("/keys").join("sefs").join("lower").display()
+            );
+            nix::mount::mount(
+                Some(source),
+                mount_path,
+                Some(fs_type.as_str()),
+                flags,
+                Some(options.as_str()),
```
mythi, line 53: we could use `dir=/keys/sefs/lower` directly?

src/runtime-boot/init/src/main.rs
```diff
@@ -4,74 +4,96 @@ extern crate serde_json;
...
+            )
+            .unwrap_or_else(|err| {
+                eprintln!("Error reading file: {}", err);
```
mythi, line 56:
```suggestion
                eprintln!("Error mounting keys: {}", err);
```

src/runtime-boot/init/src/main.rs
```diff
@@ -4,74 +4,96 @@ extern crate serde_json;
...
+            });
+
+            let KEY_FILE: &str = "/tmp/key.txt";
+            // Get the key of FS image
+            let key = {
+                let key_str = load_key(KEY_FILE)?;
+                let mut key: sgx_key_128bit_t = Default::default();
+                parse_str_to_bytes(&key_str, &mut key)?;
+                Some(key)
+            };
```
mythi, line 66: `umount` after reading

src/runtime-boot/init/src/main.rs
piotrpalcz: Done

src/runtime-boot/init/src/main.rs
piotrpalcz: Done

src/runtime-boot/init/src/main.rs
piotrpalcz: Done

src/runtime-boot/init/src/main.rs
piotrpalcz: Done

mythi (approved):

src/runtime-boot/init/src/main.rs
```diff
@@ -4,74 +4,90 @@ extern crate serde_json;
...
+            nix::mount::umount(mount_path)?;
+            let key_ptr = key
+                .as_ref()
+                .map(|key| key as *const sgx_key_128bit_t)
+                .unwrap_or(std::ptr::null());
+
+            // Example envs. must end with null
+            let env1 = CString::new("TEST=1234").unwrap();
+            let envp = [env1.as_ptr(), std::ptr::null()];
```
dcmiddle, line 69: Looks like this was in the original code, but maybe we shouldn't keep it. Do we need some dummy environment variable in order to essentially have an empty list?

mythi: It follows the Occlum "example" for how envs are passed. Back in 3d2ca1b4970cb749936ea9070bf7748267b1a490 I added it so that we get to test the functionality but that testing is still missing. I guess we could drop it but my preference is to do the cleanup in a separate PR.

dcmiddle: ok. that makes sense to me.

dcmiddle (approved):

mythi (approved):

PR #314, opened 2023-12-20
CI&build updates 

mythi:


dcmiddle (approved):

dcmiddle:
Curious why you are dropping indirect?
It looks like https://github.com/confidential-containers/enclave-cc/pull/316 has a couple indirect bumps.

mythi:
> Curious why you are dropping indirect? It looks like #316 has a couple indirect bumps.

To prioritize direct dependencies. My observation: we seem to be getting one PR a day but they've mostly been indirect deps. Given the threshold 1 open PR at a time I need to keep the queue empty on a daily basis :-) If I stop merging, we might miss some important direct deps update.

PR #319, opened 2024-01-04
bump: image-rs to new version

Xynnn007:
in old image-rs we use Attestation Agent as a lib inside to request for keys. After some refactoring, AA remains only the attestation parts. This patch updates image-rs version after that refactoring.

see https://github.com/confidential-containers/guest-components/issues/412

### DIFF
--- a/src/enclave-agent/Cargo.lock
+++ b/src/enclave-agent/Cargo.lock
@@ -174,28 +174,9 @@ dependencies = [
 ]
 
 [[package]]
-name = "attestation_agent"
-version = "0.1.0"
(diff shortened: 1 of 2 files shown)


mythi:
> We can change the rev to `confidential-containers` after merging [confidential-containers/guest-components#427](https://github.com/confidential-containers/guest-components/pull/427)

Do you think we could just wait that merge first and then update here? i.e., not merge this right now

Xynnn007:
> > We can change the rev to `confidential-containers` after merging [confidential-containers/guest-components#427](https://github.com/confidential-containers/guest-components/pull/427)
> 
> Do you think we could just wait that merge first and then update here? i.e., not merge this right now

Yes. I meant this.

Xynnn007:
Updated the git rev to `confidential-contaienrs.`

mythi (approved):

PR #414, opened 2024-06-17
ci: add clean jobs in e2e test on self-hosted runner

Xynnn007:


Xynnn007:
Let's trigger this test multiple times to check if it is idempotent

Xynnn007:
After 3 times re-triggered they all passed. I think it is good now. Let's get this merged first and then can rebase under https://github.com/confidential-containers/enclave-cc/pull/412 is

mythi:
What is wrong with the existing cleanup method?

Xynnn007:
> What is wrong with the existing cleanup method?

Existing cleanup steps are set for repo runner I think (I do not have permission to check for `Environment=ACTIONS_RUNNER_HOOK_JOB_STARTED`). For organization runner, we need to set another one. Similar to [guest-components](https://github.com/confidential-containers/guest-components/pull/583/files#diff-9313799a136638a8a26fbb8a7c114f9214c31b6cee93ffbb0b847889fbec9610R38-R43) does, we are keeping a script on the runner. This useful when different repos are using a same runner.

dcmiddle:
Thanks for the improved cleanup. I'm curious what is in post_action.sh and whether that should be in the commit? Like is there a risk of different self-hosted environments having spurious failures because of different cleanup?

docs/ci-setup.md
```diff
@@ -58,21 +58,15 @@ mkdir -p data/kbs-storage/default/image-kek
 echo LieOhvkqFcGMzZrVzt6vPWlj/F/bgYMNe45vhQpdxAA= | base64 -d > data/kbs-storage/default/image-kek/11032d96-dccd-46a3-9244-93644d76745f
 ```
 # Github Runner Service
-For `enclave-cc` e2e tests, we run a "job-started" pre-cleanup job configured
-for the runner:
-
-```
-Environment=ACTIONS_RUNNER_HOOK_JOB_STARTED=<path/to>/job-started-cleanup.sh
-```
+For `enclave-cc` e2e tests, we run a "post-action" job to clean the useless files generated by the CI.
```
mythi, line 61: I don't remember why I used `ACTIONS_RUNNER_HOOK_JOB_STARTED` but if it needs to be a "post-action", then the built-in mechanism via `ACTIONS_RUNNER_HOOK_JOB_COMPLETED` is preferred rather than using `e2e.yaml` for this.

Xynnn007:
@dcmiddle yes I have updated the content of the clean scripts parts in `docs/ci-setup`. The other parts are only sub-command identifier. Because this script is also used by `guest-components`, thus I think we do not need to add the whole content for that, but only enclave-cc parts.

docs/ci-setup.md
Xynnn007: Yes. I think post-action should be better. This would make every ci pipeline idempotent, without leaving anything left on the machine.

Xynnn007: I think this would let all jobs on the runner to run this clean up script. this is not expected. As the runner is shared among multiple repos, different repo would have different steps of clean jobs.

mythi: That would be my expectation and that's what we'd end up doing with this change also: both `guest-components` and `enclave-cc` end up running `${HOME}/script/post_action.sh`. `GITHUB_REPOSITORY` variable can be used to run repo specific commands.

Xynnn007: Ok I think it deserves a try.

Xynnn007:
I think the CI works now. I only changed a little about the ci script due to my taste : )

mythi (approved):

PR #442, opened 2024-08-22
init: fix syscall parameters

mythi:
PR #270 combined the boot-instance and agent-instance in one but the modifications to "init" missed some syscall() parameters to Occlum.

SYS_MOUNT_FS takes two parameters so add the missing one.


mythi:
@qzheng527 PTAL and perhaps approve

PR #449, opened 2024-09-02
workflow updates

mythi:
Closes: #454

### DIFF
--- a/.github/workflows/basic.yaml
+++ b/.github/workflows/basic.yaml
@@ -7,6 +7,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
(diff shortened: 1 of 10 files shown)


dcmiddle:
> ci: use pinned action versions and change rust-toolchain action

Can you say why you want to use checksums rather than versions?

mythi:
> > ci: use pinned action versions and change rust-toolchain action
> 
> Can you say why you want to use checksums rather than versions?

I've had to explain this in other PRs too so I'll just add a link :-) https://github.com/confidential-containers/operator/pull/435#discussion_r1744826698

dcmiddle (approved):

PR #516, opened 2025-10-16
Announce proposed deprecation

dcmiddle:


Xynnn007:
The doc link is now removed (as CCv0 branch is removed).

mythi (approved):

mythi:
> The doc link is now removed (as CCv0 branch is removed).

thanks for checking the error. I'd think we can just ignore it in this PR

